PR #18, opened 2015-01-13
Compilable / usable on Windows: removed gitty dependency, added execSync

duizendnegen:


LevelbossMike:
Thank you! This seems like a good change. Can you please update the necessary tests so the build is passing again?

I think it would be a good Idea to wrap the git interaction in a wrapper object so this functionality is more stable to changes and changes to the underlying calls don't break the tests as easily.

duizendnegen:
Yep!
Wrapping the git interaction in a wrapper could be added in later, I wouldn't mind doing that myself; I'm mostly focusing on getting the Azure adapters in (the index adapter is completed, have to write a way of fetching its contents in C# now)

LevelbossMike:
Great! I am happy to merge this. Could you please combine this into one commit before though?

duizendnegen:
Will open a new PR, messed some stuff up with squashing.

PR #20, opened 2015-01-14
use redis with secret

philipheinser:
Added an example of how to use an reds-db with secret to readme. Took me a while to figure this out.


LevelbossMike:
Thanks a lot!

Issue #21, opened 2015-01-14
Deploying environments with different configurations


tylr:
We deploy several environments for our app. They are configured within `config/environment.js`.

Currently the ember-deploy script is explicit about building the application with the production environment.

See: https://github.com/LevelbossMike/ember-deploy/blob/master/lib/commands/deploy.js#L22

This has the side-effect of not respecting all the env vars that my application needs to be aware of for each environment. e.g. apiHost, websocketHost, googleAnalyticsId, etc.

I considered adding these configurations to the deploy.json, but it seems a little out of the scope of that files purpose, or require a refactor that ends up conflating my application configuration with the deploy process.

Thoughts?

LevelbossMike:
Hi and thanks for opening this issue!

Good catch and a useful feature request. It would be pretty easy to refactor this so that

`ember deploy --environment staging`

will build the application by passing the `staging` environment instead of `production` hard-coded. Most likely every environment except of development should be respected in the command. Building and uploading a development build does not seem very useful to me but it might be to some people.

Are you interested in sending a PR for this? I will have time to do this on the weekend probably but if you need this right now it's pretty easy to implement.

Until now I built the app with the respective environment myself and then used the `ember deploy:index` and `ember deploy:assets` tasks manually to upload necessary files. But including this in `ember-deploy` directly seems like a much more cleaner solution.

tylr:
Great, that provides the clarity I needed. I've already got something working. I'll clean it up, fork and make a PR before the weekend.

We're migrating the bustle.com CMS to the ember-cli, deployment is the last step. :)

LevelbossMike:
Bustle.com? Nice! Let me know if I can be of any more help if you run into problems.

tylr:
Closing after agreeing on an approach in #23

PR #23, opened 2015-01-15
Pass environment down to build task

tylr:
A fix for #21

There might be other considerations to make before merging this, but it works for my use case.

This will default to a development build. This may not be ideal, but it seemed better than making any other assumptions. And defaulting to `production` seems like a bad idea.

Another idea, is that we provide a convention in `deploy.json` to pass build properties to the build task. That would potentially decouple the deploy environment from the ember build environment. e.g. `deploy.json` could look like:
```
{
  staging: {
    build: {
      assetHost: 'cdn.bustle.com',
      apiHost: 'staging.api.bustle.com'
    }
  }
}
```
Does that make sense?

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,7 @@ If you don't install the redis- and s3-adapters you will need to use a custom ad
   },
 
   "staging": {
+    "buildEnv": "staging", // Override the environment passed to the ember asset build. Defaults to 'production'
     "store": {
       "host": "staging-redis.example.com",
       "port": 6379
--- a/deploy.json
+++ b/deploy.json
@@ -12,6 +12,7 @@
   },
 
   "staging": {
+    "buildEnv": "staging",
     "store": {
       "host": "staging-redis.firstiwaslike.com",
       "port": 6379
--- a/lib/commands/deploy.js
+++ b/lib/commands/deploy.js
@@ -1,3 +1,5 @@
+var ConfigurationReader = require('../../utilities/configuration-reader');
+
 module.exports = {
   name: 'deploy',
   description: 'Deploys an ember-cli app',
@@ -12,6 +14,9 @@ module.exports = {
     var AssetsTask      = require('../tasks/assets');
     var BuildTask       = this.tasks.Build;
 
+    var config = new ConfigurationReader({
+      environment: commandOptions.environment
+    });
     var buildTask = new BuildTask({
       ui: this.ui,
       analytics: this.analytics,
@@ -19,7 +24,7 @@ module.exports = {
 
     });
     var buildOptions = {
-      environment: "production",
+      environment: config.buildEnv || 'production',
       outputPath: "dist/",
       watch: false,
       disableAnalytics: false
--- a/utilities/configuration-reader.js
+++ b/utilities/configuration-reader.js
@@ -16,6 +16,7 @@ function ConfigurationReader(options) {
 
   this.store = this._config[this._environment].store;
   this.assets = this._config[this._environment].assets;
+  this.buildEnv = this._config[this._environment].buildEnv || 'production';
 }
 
 module.exports = ConfigurationReader;


LevelbossMike:
Thanks for your pull request!

Defaulting to development seems kind of strange because it does not make much sense to deploy the development version anywhere in my opinion. If you wanted to show your development version to somebody you can always send your network ip and people will be able to connect to your development machine (interested people will most likely be on your corporate network or you can use a service like https://ngrok.com/).

All other environments make more sense because you probably will have some notion of production like environments you want your QA-departement or similiar people to test against. I suppose a 'sane' default flow would be as easy as adding a function that checks if the passed environment is development then uses production and in other cases just uses the passed environment. Of course this feels kind of hacky but well..

```js
var sanitizeBuildEnvironmentOption = function(environmentName) {
  if (environmentName === 'development') { return 'production'; }
  return environmentName;
}
```

I really can't think of a situation where I wanted to deploy the development environment anywhere. Of course defaulting to production does not make much sense either but if you really have a small project where you just want to test out if the deploy will work this would at least not do 'the wrong thing'.

I also would back away from moving customization for builds into the `deploy.json` file because ember-cli already has the necessary tools to customize builds (`environment.js`) and I don't think it is a useful approach to duplicate this.

What's your opinion on this?

tylr:
I'm less concerned with why you would want to deploy development environment, and more concerned with defaulting to production being a bad idea.

What if your production build points to the production apiHost? Now I have a build not intended for production doing potentially destructive things against production data. In general, nothing should ever default to production.

I also think there _is_ a usecase for defaulting to development. It's already proven useful to do a dev asset build and test the ember app in my host application. (e.g. rails, express...). 

I may very well be over thinking it, but if we default to `production` I'll have to patch it anyway, because that is potentially harmful.

LevelbossMike:
You should not be able to "destroy" data on production without valid credentials (assuming you use different credentials in dev than in prod) but yeah, development setup will be different across every team and that is what I meant by defaulting to `production` being the wrong thing unless you are deploying the most trivial application. API-Host will be the wrong one, you will run into cross-site issues etc.

We could also just remove the default environment for the deploy command and error out with a useful error message if no environment is passed to `ember deploy`?! Does not demo that well but may be the right thing to do anyway.

I think there are two alternatives:

A) Error-out
B) Default to `production` with an option in `deploy.json` to turn this behaviour off

B is mostly a 'demos' well option. I already gave a talk about ember-deploy where I said that `ember-deploy` defaults to development (maybe some people rely on this?!) and also it is easier to test out stuff quickly without the need to pass an environment.

What is your preferred solution?

tylr:
I hacked it together so that `ember deploy` defaults to `staging`. Since that's ultimately what the developers on our team intend to do when they run `ember deploy`.

What if `deploy.json` had a configuration paramater called `buildENV` that it passed to the asset build, and default it to `production`? That approach seems demo friendly, makes the most common assumptions, and lets you configure the desired behavior.

LevelbossMike:
The buildENV idea sounds great!

LevelbossMike:
@tylr Do you plan to integrate your suggested solution? Otherwise I will tackle this myself but I'd be more than happy to accept your page with buildENV functionality.

tylr:
@LevelbossMike I'm back on it today. Will follow up with a PR.

tylr:
@LevelbossMike feel free to review / provide any suggestions.

LevelbossMike:
Great work! looks good to me. It would be really great if you also could mention this in the README and rebase so we don't get this merge commit. 

Really excited for this change!

tylr:
@LevelbossMike updated readme and squashed. Could probably stand to break out and document the configuration options, but I think this will suffice for the PR.

LevelbossMike:
This is awesome! Thank you!

PR #24, opened 2015-01-19
added nodejs example

philipheinser:


LevelbossMike:
This is great! Can you please add a small description to this part of the README? I had a look at the README with your patch applied and it is easy to overlook if you only add the link.

philipheinser:
@LevelbossMike like this?

LevelbossMike:
Thx a lot!

PR #25, opened 2015-01-19
Allow custom config file

jbescoyez:
This PR adds the `--config-file` option to every command. It allows to store the deploy config in another location than `deploy.json`. Moreover, it allows to read the config from a `.js` file and, by extension, to take advantage of environment variables.


LevelbossMike:
Nice! Thank you very much! Can you please also update the README to introduce users to this change? Especially an example of the improvement that shows how one would use a js file would be great.

LevelbossMike:
merged in a separate pull request. thanks for your contribution :+1:

PR #27, opened 2015-01-21
Don't GZIP when specified it is not required (not all deployments like G...

duizendnegen:
...ZIP content)


LevelbossMike:
Thank you!

Issue #31, opened 2015-02-02
Cannot to install module with node 0.11.16, execSync dependency fails to build


kriswill:
While attempting to install using node 0.11.16, npm 2.4.
```sh
→ node -p process.versions
{ http_parser: '2.3',
  node: '0.11.16',
  v8: '3.28.73',
  uv: '1.0.2',
  zlib: '1.2.8',
  modules: '14',
  openssl: '1.0.1l' }
```
```sh
→ npm --version
2.4.1
```
```sh
→ ember install:addon ember-deploy
version: 0.1.12
Installing packages for tooling via npm..[execsync v1.0.2] Attempting to compile native extensions.
Installing packages for tooling via npm[execSync v1.0.2]
    Native code compile failed!!
Installed packages for tooling via npm.
Installed addon package.
```

kriswill:
BTW, this is also broken in v0.12, the new stable node.

LevelbossMike:
did your pr fix this? Or do we need to do more work to geht this fixed? I am on node 0.10.x

kriswill:
Yes, the PR does fix it.  When are you publishing a new version?

LevelbossMike:
just published 0.3.1. Sorry for the inconvenience and thx for your fix.

kriswill:
something else in this version is causing fingerprinting to fail.  I'll investigate.

kriswill:
Okay, so the buildEnv wasn't working before, but now it is.  so I had to add this to my Brocfile.js to get it to enable fingerprinting:

```js
if (app.env === 'staging') app.options.fingerprint.enabled = true
```

LevelbossMike:
fingerprinting is actually done by ember-cli itself. You will have to enable it for every build env you want to use it except for production (production has fingerprinting enabled by default). That was the case the last time I checked. Can you confirm this behaviour?

i.e. you could do something like this:

```javascript
//Brocfile.js
isProductionLikeBuild = ['production', 'staging'].indexOf(process.env.EMBER_ENV) > -1;

var app = new EmberApp({
  fingerprint: {
    enabled: isProductionLikeBuild,
    prepend: <your cdn url>
  }
});

//...
```

PR #32, opened 2015-02-03
add support for execSync in node >= 0.11

kriswill:
Addresses #31


kriswill:
I had to refactor a little to make the sinon stub work, since the return values are different from the execSync polyfill and the actual node 0.11 child_process.execSync functions.

LevelbossMike:
Thx!

PR #33, opened 2015-02-20
Link to new S3 Index Adapter

pootsbook:
Serve the `index.html` file out of S3 directly or use S3 as storage behind a proxy instead of Redis. Addon written by @Kerry350, feedback welcome!

(Also minor edit: use the scheme for S3 link, and name S3 with a capital S.)


Kerry350:
Thanks @pootsbook :) I've got a detailed blog coming on how you would use this adapter - as it would require an extra bucket for previews (but not revision activations). That'll be slotted in to the README real soon, the adapter itself is working though with all of the public methods. Great addon by the way, was easy to work with :)

LevelbossMike:
Thanks to both of you! Great addon!

PR #34, opened 2015-02-22
Allow defining of tagging adapter in deploy.json

jamesfid:
In writing a custom `tagging-adapter`, I noticed that the `tagging` configuration value (referenced in [lib/tasks/deploy-index.js](https://github.com/LevelbossMike/ember-deploy/blob/master/lib/tasks/deploy-index.js#L18)) never actually gets set by the `configuration-reader`.  This PR should fix the issue as well as show in the README an example of defining a custom `tagging-adapter` for usage.

### Questions
One issue that may need to be addressed is where the defaulting of configuration values should be done.  Presently I have the `tagging-adapter` defaulted to `sha` in the `configuration-reader` itself.  This defaulting is duplicated in [lib/tasks/deploy-index.js](https://github.com/LevelbossMike/ember-deploy/blob/master/lib/tasks/deploy-index.js#L18).  

What is the preferred pattern for defaulting configuration values?  I can adjust the PR accordingly once this is defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,6 +218,8 @@ The following adapters have already been developed:
 * [ember-deploy-s3](https://github.com/LevelbossMike/ember-deploy-s3), assets-adapter for AWS [S3](http://aws.amazon.com/s3/)
 * [ember-deploy-s3-index](https://github.com/Kerry350/ember-deploy-s3-index), index-adapter for AWS [S3](http://aws.amazon.com/s3/)
 * [ember-deploy-azure](https://github.com/duizendnegen/ember-deploy-azure), index-adapter and assets-adapter for Azure Tables & Blob Storage respectively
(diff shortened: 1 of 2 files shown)


LevelbossMike:
Thx for this PR. I will not have time to look into this during the day but I will come back to you asap. Defaulting configuration options is somewhat cumbersome right now but the ideal way would be that configuration reader knows about the default values and initializes a default configuration when some things are missing after reading from deploy.json. this is definitely the next thing that should be improved in `ember-deploy`.

LevelbossMike:
There's already an issue for the configuration reader https://github.com/LevelbossMike/ember-deploy/issues/29 . If you have the time to tackle this this would be an awesome contribution. I am pretty busy right now so I don't know when I will find the time to add this myself.

jamesfid:
This is no longer valid per merging of #37

PR #36, opened 2015-02-24
Allow custom config file via --deploy-config-file.

lukemelia:
Based on work from jbescoyez in #25, which this PR replaces.


LevelbossMike:
:+1: This is great! Thanks a lot!

jbescoyez:
@lukemelia much better. Tnx!

PR #39, opened 2015-02-24
Add a force option in the config which will override an already uploaded...

jbescoyez:
... index.htlm version

Depends on: https://github.com/LevelbossMike/ember-deploy-redis/pull/1

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,8 @@ By default, `ember-deploy` expects a `deploy.json` file in the root of your Embe
     "store": {
       "type": "redis", // the default store is 'redis'
       "host": "localhost",
-      "port": 6379
+      "port": 6379,
+      "force": false // when true, override the current version even if the same tag is already registered
     },
     "assets": {
     "type": "s3", // default asset-adapter is 's3'
--- a/lib/tasks/deploy-index.js
+++ b/lib/tasks/deploy-index.js
@@ -16,6 +16,7 @@ module.exports = Task.extend({
     });
 
     var adapterType = config.store.type || 'redis';
+    var forceUpload = config.store.force || false;
     var taggingAdapterType = config.tagging || 'sha';
     var manifest = this.project.name();
 
@@ -42,7 +43,7 @@ module.exports = Task.extend({
     return readFile('dist/index.html')
       .then(function(fileContent) {
         ui.writeLine(chalk.blue('\nTrying to upload `dist/index.html`...\n'));
-        return deploy.upload(fileContent);
+        return deploy.upload(fileContent, forceUpload);
       });
   }
 });


LevelbossMike:
Mmh not sure about this. I get why this is useful when developing ember-deploy itself but I am not sure if this is a useful setting when deploying an application. Do you have a use case for this in your app?

jbescoyez:
Here is my story :) - I use Codeship for code deployment. For every environment (staging, prod, ...) I have specific environment variables defined for configuration. When I change environment variables, I would like to re-upload a new build of my ember app. Since the commit has not changed. I'm not allowed to that. The current workaround I use is to push an empty commit (ugly).

jbescoyez:
@LevelbossMike WDYT about this use case?

achambers:
I think I tend to agree with @LevelbossMike.  While your solution @jbescoyez would work, I don't feel like it's the right one as I don't think it is fixing the real problem.  The real problem is that the git sha, while it is unique enough, it doesn't fully reflect the state of an app, because, as you said config can change without the git commit hash changing.

Maybe, the real fix is to use a different value for the upload key.  By default, the config is stored in a meta tag in the index html.  So maybe a fingerprint of the index.html file is the way to go instead?

achambers:
Not to mention, people not using git (maybe mercurial) won't be able to use this as is. So maybe that's another case for not using the git hash too.

achambers:
Having said all that, I do like the idea of being able to deploy where the key is actually the branch name which is much easier to remember. In which case you **would** want to over write the index.html because the state of the branch changes. So there is that too. 

I guess my point is, @jbescoyez, there is more discussion that needs to happen around this area.

jamesfid:
The ability to name a revision after a branch, or maybe a run option, would be nice.  In our case, we deploy an early build of the client to certain users under the revision key of 'beta' (from branch `beta`).  It would be nice if something like this was support as opposed to manually doing it.

Switching gears, using the git hash was not a great solution for our setup either.  We use `npm link` to symlink in repositories stored elsewhere on the machine.  When we update our client code, the base project's git repo has no changes to be made, so originally we were doing the blank commit approach to solve this as well.  As a solution, I wrote a quick `TaggingAdapter` to SHA the index.html itself and use that, similar to what @achambers noted above.  It's simple, but you can see it at [ember-deploy-tagging-index-sha](https://github.com/jamesfid/ember-deploy-tagging-index-sha)

Anyway, just my two cents.  I look forward to more discussion.

lukemelia:
SHAing the index.html contents is a pretty good solution, since it should include the SHAs of the JS and CSS dependencies, which in turn include SHAs of images being deployed. Another possibility is to generate a manifest of all files to be deployed that can be used for a faster differential deploy of assets (to S3, e.g.) and use the SHA of the manifest as the revision ID.

LevelbossMike:
This is definitely something we can improve but just forcing uploads seems like a hack to me. I am with @achambers here and think that the naming should be improved. If ember-cli-deploy provided a way to 'tag'-releases by hand (maybe just appending a version number that the user provides) this would work for you @jbescoyez correct?

danshultz:
I really like @lukemelia solution to this of using the SHA of the index.html page. This will then reflect the embeded config (different per env) as well any variance in references of the assets.

@jbescoyez - In the scenario with codeship, how do you differentiate between the index.html files between environments if a force upload is used?

jbescoyez:
@LevelbossMike Actually, I'm not sure to understand the use case for *not* updating the `index.html` when the key already exist.

@danshultz For the moment, I use a different Redis store for staging and production which is not optimal. However, I think it is a different issue. We might add a key differentiator in the `store` configuration.

@lukemelia SHAing the `index.html` might be a solution. However, it will not allow to load an arbitrary older version of your index.html for testing purposes. At the moment, I can test different branches based on the git revision id of there last commit.

LevelbossMike:
@jbescoyez This provides a level of security when deploying index.html manually with non-commited changes. (i.e. current sha is deployed, somebody creates a build, deploys and overrides the currently deployed revision with an index.html that references non-existend assets)

jbescoyez:
@LevelbossMike Ok. I see. I manage that by using different environment variables and stores for development, staging and production but I understand there is a need for that.

I think the @lukemelia solution seems the best to meet both use cases. I guess I can do that with a tagging adapter. I can develop a new "sha-index" tagging adapter inspired by https://github.com/ember-cli/ember-cli-deploy/blob/master/lib/utilities/tagging/sha.js . Make sense?

lukemelia:
@jbescoyez a git-sha based plugin seems like a great idea. Would love your help in #47 thinking about what hooks we will need in the 0.5.0 plan to support this.

jbescoyez:
@lukemelia Ok, I'm going to have a look at #47 and see how can I help.

devinus:
@lukemelia Is there any chance we can support a `--force` option in the interim? It's a very quick fix that supports the deployment pipeline for a lot of us.

devinus:
@jbescoyez We also use Codeship. What did you guys come up with or are you still struggling with this?

achambers:
@devinus What version of ember-cli-deploy are you using? The ember-cli-deploy-redis plugin for ecd v 0.5.0 supports this functionality already.

https://github.com/ember-cli-deploy/ember-cli-deploy-redis#allowoverwrite

devinus:
@achambers I'm using https://github.com/Kerry350/ember-deploy-s3-index but I was under the impression that it wasn't up to the index adapter?

achambers:
@devinus You are using ember-cli-deploy v0.4.x by the looks of it. 0.5.0 has made a large architectural shift which allows for much more flexibility. It will also allow you to do what you have been requesting.

I did a talk about v0.5.0 which should hopefully make things clearer for you -> https://vimeo.com/139125310

PR #40, opened 2015-02-24
Removed erroneous conflict markers.

jamesfid:
Quick fix to remove some conflict markers causing issues when calling `uploadAssets()`.


LevelbossMike:
:+1:

PR #56, opened 2015-02-28
Deprecated commands no longer needed.

### LINKED ISSUE #50: Add deprecation warnings for former ember-cli-deploy commands
For users of the former ember-cli-deploy project that have upgraded to the new version without being aware of the changes, we should add some deprecation warnings for when they run commands from the former project that no longer exist in this new one.

The commands that will print deprecation warnings are:

- [ ] ember deploy:assets
- [ ] ember deploy:index
- [ ] ember activate
- [ ] ember deploy:versions

achambers:
Deprecated the following commands:

- ember deploy:index
- ember deploy:assets

As well as commands from the former ember-cli-project:

- ember activate
- ember deploy:versions

Closes #50

### DIFF
--- a/lib/commands/deprecated/deprecated-command.js
+++ b/lib/commands/deprecated/deprecated-command.js
@@ -1,0 +1,12 @@
+var chalk       = require('chalk');
+var SilentError = require('ember-cli/lib/errors/silent');
+
+module.exports = function(command) {
+  var description = command.description || '';
+  command.description = chalk.red('[DEPRECATED] ') + description;
+  command.run = function() {
+    throw new SilentError('This command has been deprecated.  Please see `https://github.com/ember-cli/ember-cli-deploy` for details.');
+  }
+
+  return command;
+};
--- a/lib/commands/deprecated/activate.js
+++ b/lib/commands/deprecated/activate.js
@@ -1,0 +1,15 @@
+var DeprecatedCommand = require('./deprecated-command');
+
+module.exports = DeprecatedCommand({
+  name: 'activate',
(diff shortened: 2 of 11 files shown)


lib/commands/deprecated/deprecated-command.js
```diff
@@ -0,0 +1,12 @@
+var chalk       = require('chalk');
+var SilentError = require('ember-cli/lib/errors/silent');
+
+module.exports = function(command) {
+  var description = command.description || '';
+  command.description = chalk.red('[DEPRECATED] ') + description;
+  command.run = function() {
+    throw new SilentError('This command has been deprecated.  Please see `https://github.com/ember-cli/ember-cli-deploy` for details.');
```
lukemelia: nit: "deprecated" means still works but will be removed in the future. We should probably refer to this as "no longer supported"

achambers: Ah, good point.

lukemelia:
How will this work with `ember help`? Ideally these commands would not appear. Is that possible?

achambers:
@lukemelia I modify the command descriptions to add [NO LONGER SUPPORTED] in red text.  So it looks something like this:

ember activate <key> <options...>
  [NO LONGER SUPPORTED] Activate a deployed version of index.html

I'd need to look deeper into how ember-cli creates the commands but I imagine, to remove them from ember help would mean that they can't be executed.  Which kind of defeats the purpose of adding the message in the first place.

lukemelia:
@achambers sounds good. doesn't look like ember-cli has a way to hide a command from help.

PR #58, opened 2015-02-28
Add blueprint to auto generate config/deploy.js

### LINKED ISSUE #52: Add blueprint to install starter deploy.js on install
When ember-cli runs `ember install:addon ember-cli-deploy`, a blueprint should run to output a starter `config/deploy.js` file. The file should have some commented examples showing users how switch based on deploy environment, how to use environment variables, and execute a command synchronously to get a value.

achambers:
* Adds `deploy-config` blueprint to `ember generate`
* Populates config/deploy.js with example config

Closes #52

### DIFF
--- a/blueprints/deploy-config/files/config/deploy.js
+++ b/blueprints/deploy-config/files/config/deploy.js
@@ -1,0 +1,43 @@
+module.exports = {
+  //development: {
+    //store: {
+      //type: 'redis', // the default store is 'redis'
+      //host: 'localhost',
+      //port: 6379
+    //},
+    //assets: {
+      //type: 's3', // default asset-adapter is 's3'
+      //gzip: false, // if undefined or set to true, files are gziped
+      //gzipExtensions: ['js', 'css', 'svg'], // if undefined, js, css & svg files are gziped
+      //accessKeyId: '<your-access-key-goes-here>',
+      //secretAccessKey: process.env['AWS_ACCESS_KEY'],
+      //bucket: '<your-bucket-name>'
+    //}
+  //},
+
+  //staging: {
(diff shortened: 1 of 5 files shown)


blueprints/deploy-config/files/config/deploy.js
```diff
@@ -0,0 +1,43 @@
+module.exports = {
+  //development: {
```
achambers: Just copied the example deploy.js form the README.  We will need to remember to change this file as well as the README as the API changes

package.json
```diff
@@ -42,7 +42,8 @@
     "ember-cli-deploy"
   ],
   "ember-addon": {
-    "configPath": "tests/dummy/config"
+    "configPath": "tests/dummy/config",
+    "defaultBlueprint": "deploy-config"
```
achambers: I renamed the blueprint to `deploy-config` so the command reads `ember g deploy-config`.  I thought it was better than `ember g ember-cli-deploy`.

achambers:
Not really sure the best way to test this.  Or if it is possible, or of any value.  Thoughts anyone?

lukemelia:
@achambers there is a way to make a blueprint run automatically as part of ember install:addon I'm out right now but you should be able to find an example in sam's pretenderify project

achambers:
@lukemelia The `"defaultBlueprint": "deploy-config"` line in the package.json should do that.  According to the [ember-cli docs](http://www.ember-cli.com/#configuring-your-ember-addon-properties) anyway.

lukemelia:
@achambers awesome, that's what I get for trying to read code on my phone!

PR #59, opened 2015-02-28
[WIP] Revamp README

### LINKED ISSUE #49: Update README to reflect move to ember-cli/ember-cli-deploy
Since moving `ember-deploy` to `ember-cli/ember-cli-deploy` we need to update the README.  It probably needs quite an overhaul but below are a few of the things mentioned in Slack.

- [ ] Change references to `ember-deploy` and LevelBossMike
- [ ] De-emphasise references to Luke's 'Lightening Deployments' presentation
- [ ] Add something to the top of the README to briefly explain the migration that has happened and why users of the former ember-cli-deploy will probably be experiencing issues.

achambers:
**Please do not merge this**. It is a work in progress.

Please feel free to comment on or modify anything you think makes sense and adds value.

Closes #49


README.md
```diff
@@ -1,279 +1,43 @@
...
-## Config file
+Alternatively, you can specify a different file location by passing the `deploy-config-file` option to the `ember deploy` command.
 
-By default, `ember-deploy` expects a `deploy.js` file in the `config/` directory of your Ember-CLI project root. In this file you tell `ember-deploy` about the necessary credentials for your file hoster and for your key-value store. An example could look like this:
+An example of the configuration file is as follows:
 
-```js
+```javascript
+// config/deploy.js
 module.exports = {
```
achambers: Have left the config file example blank until we have a better idea of what it will actually look like.  This is not clear in my head at the moment as I believe it's going to change quite drastically with the idea of the pipeline and plugins.

README.md
```diff
@@ -1,279 +1,43 @@
...
-
-Ember deploy will gzip your `js` and `css` assets and copy these gzipped assets and all other assets from ember-cli's `dist`-folder to the `tmp/assets-sync`-directory. You can do whatever you want in an `asset-adapter`'s `upload`-method but most likely you will upload these assets to some kind of static-asset server with the `upload`-method.
-
-This method has to return a `RSVP.Promise`.
-
-### Tagging-Adapters
+## Commands
 
-`tagging-adapters` take care of tagging revisions when uploading revisions via the `deploy-index`-command. `ember-deploy` comes bundled with the `sha`-tagging-adapter that will tag your revisions according to the current git-sha.
+ember-cli-deploy adds the following deployment commands to Ember CLI:
```
achambers: Have only added `ember deploy` at this stage as I'm unclear as to what is actually needed other than that.  How will `ember activate` actually fit in if it's is really only relevant to one potential deployment strategy (the lightening method).

README.md
```diff
@@ -1,279 +1,43 @@
...
 
-Tagging Adapters need to implement the following method for ember-deploy to know what to do:
+- `ember deploy` Initiates the deployment pipeline, invoking any registered ember-cli-deploy plugins at the defined hook points
```
achambers: Have only added `ember deploy` at this stage as I'm unclear as to what is actually needed other than that. How will ember activate actually fit in if it's is really only relevant to one potential deployment strategy (the lightening method).

achambers:
Lads, here's my first crack at a new README.  Feels like there are still a few things up in the air but it's a start.  Please leave feedback.

Questions and comments that I have currently are:

1. Have left the example `config/deploy.js` blank as I don't feel what we currently have is a good reflection of where we are going.
2. Have only added `ember deploy` at this stage as I'm unclear as to what will actually be needed other than that.  `ember activate` feels very specific to the lightening deployment approach.  Happy to add it back in if need be.
3. I have taken the detail on what a plugin hook should resolve/reject with from the [ember-cli ARCHITECTURE.md](https://github.com/ember-cli/ember-cli/blob/master/ARCHITECTURE.md) file.  It sounded reasonable.  Let me know if it's not.
4. Will we provide a `Plugin` object that can be extended?  This seems like it could be a good idea.
5. Following on from number 4 above, will we instantiate the plugin with a UI instance to allow it to write to the console?  This is how Tasks and Commands are created and seems like a good idea for us.
6. Maybe it could be a good idea for us to have a PLUGINS.md or PLUGIN_HOOKS.md file in the root which gives a detailed overview of what makes up a plugin and what the hooks are for.  Especially, if the list of hooks grows this might become useful.  And it means we can keep the README to be more about what ember-cli-deploy is about generally.

Anyway, that is a bit of a brain dump.  I welcome any feedback.

README.md
```diff
@@ -1,279 +1,106 @@
...
-      // you can add multiple adapters in one ember-deploy-addon
-    }
-  };
-}
+```javascript
+// plugins/lightening-fast-deployment.js
 
-module.exports = EmberDeploySuperAwesome;
-```
+var Plugin = require('ember-cli-deploy/models/plugin');
```
danshultz: There was a brief discussion about not having the plugins depend explicitly and the possibility of "...plugins implementing a willDeploy hook that essentially runs validation: e.g. is required  configuration present? that could include checking for a minimum version of ember-cli-deploy"

I did run into an issue yesterday where a plugin depended on ember-deploy and it stomped on ember-cli-deploy that I had installed. Not sure that would be an issue though with ember-cli-deploy in the future. Also, this potentially could break a plugin due to a version mismatch in the future or a weird cross dependency issue.

I mentioned peerDependencies but that could also be a can of worms...hence the thought of checking the version in a hook or something.

lukemelia:
Looking good so far. Some of this could end up living in DEVELOPING_PLUGINS.md per #71. I'm going to tag this for 0.5.0.

lukemelia:
Closing in favor of all the awesome docs work that has landed.

PR #60, opened 2015-03-01
update ember-deploy references to be ember-cli-deploy

danshultz:
Alright, I think I covered getting all of these renamed properly. I left the readme and the changelog be as I expect these to get their own overhaul.

### DIFF
--- a/utilities/adapter-registry.js
+++ b/utilities/adapter-registry.js
@@ -5,7 +5,7 @@ var merge          = require('lodash-node/modern/objects/merge');
 
 module.exports = CoreObject.extend({
   init: function() {
-    if (!this.project) { return };
+    if (!this.project) { return; }
 
     this.project.addons.forEach(this._mergeDeployAdapters.bind(this));
   },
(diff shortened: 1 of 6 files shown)


utilities/adapter-registry.js
```diff
@@ -33,7 +33,7 @@ module.exports = CoreObject.extend({
   },
 
   _mergeDeployAdapters: function(addon) {
-    if (addon.type === 'ember-deploy-addon') {
+    if (addon.type === 'ember-cli-deploy-addon') {
```
lukemelia: This is a breaking change to plugins, and we plan a different breaking API change (require package.json keywords to include "ember-cli-deploy-plugin" and implement hooks instead of adapters hash). Maybe we should wait to break the plugins until we have the planned API ready for them?

danshultz: Yes, that would make sense to wait to change this I think. Most plugins should just work with holding off to make this change unless the plugin depended on `ember-deploy` directly.

Reverted d897962

PR #63, opened 2015-03-01
Fix regression to the type of object that was being passed to assets adapters as “config”.

### LINKED ISSUE #62: Regression in how config is exposed to adapters
via @danshultz "Just noticed the following when testing - it appears that encapsulating the defaults (https://github.com/ember-cli/ember-cli-deploy/commit/d838cd750c274f25d041c8ef7b3315f1cac7bf9c) seems to cause a breakage here --> https://github.com/ember-cli/ember-cli-deploy/blob/master/lib/tasks/assets.js#L59"

lukemelia:
Fix regression to the type of object that was being passed to assets adapters as “config”. Closes #62

* Slight refactoring to AssetsTask to make it’s interaction with AssetsUploader more testable
* this change means we are not being as lazy with default values as before. That’s ok.
* ConfigurationReader now needs to be constructed with a project reference

### DIFF
--- a/lib/commands/deploy.js
+++ b/lib/commands/deploy.js
@@ -1,4 +1,5 @@
 var ConfigurationReader = require('../utilities/configuration-reader');
+var AssetsUploader = require('../utilities/assets-uploader');
 
 module.exports = {
   name: 'deploy',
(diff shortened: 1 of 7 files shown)


achambers:
@lukemelia I merged PR #64 first, so might be worth a rebase of this one.

lukemelia:
Rebased

LevelbossMike:
:+1:

Issue #72, opened 2015-03-02
Revert Unsupported Commands


achambers:
For 0.4.0 we want to support any previous commands.  Therefore we should revert #56 .

PR #74, opened 2015-03-03
Revert Unsupported Commands back to Deprecated for 0.4.0 release

danshultz:
This reverts the commit that moved the original ember-cli-deploy commands to unsupported for users coming to this project from the original ember-cli-deploy package. It also lists a few original `ember-deploy` commands as deprecated also.

@achambers thanks for deprecating and then un-supporting in 2 commits. Made this an easy item to knock out this morning.

This closes issue #72

### DIFF
--- a/lib/commands/deprecated/activate.js
+++ b/lib/commands/deprecated/activate.js
@@ -1,4 +1,8 @@
+var chalk = require('chalk');
+var Promise = require('ember-cli/lib/ext/promise');
+var SilentError         = require('ember-cli/lib/errors/silent');
 var DeprecatedCommand = require('./deprecated-command');
+var ActivateCommand = require('../activate');
 
 module.exports = DeprecatedCommand({
   name: 'activate',
@@ -10,6 +14,18 @@ module.exports = DeprecatedCommand({
   ],
 
   availableOptions: [
-    { name: 'environment', type: String, default: 'development' }
-  ]
+    { name: 'environment', type: String, default: 'development' },
+    { name: 'deploy-config-file', type: String, default: 'config/deploy.js' }
+  ],
+
+  run: function (commandOptions, rawArgs) {
+    var key = rawArgs[0];
+    if(!key) {
+      var message = chalk.yellow('The `ember activate` command requires a deploy version to be specified.');
+      return Promise.reject(new SilentError(message));
+    }
+    commandOptions.revision = key;
+    return ActivateCommand.run.apply(this, arguments);
+  }
+
 });
--- a/lib/commands/deprecated/activate.js
+++ b/lib/commands/deprecated/activate.js
@@ -1,6 +1,6 @@
-var UnsupportedCommand = require('./unsupported-command');
+var DeprecatedCommand = require('./deprecated-command');
 
-module.exports = UnsupportedCommand({
+module.exports = DeprecatedCommand({
   name: 'activate',
   description: 'Activate a deployed version of index.html',
   works: 'insideProject',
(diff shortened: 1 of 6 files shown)


achambers:
@danshultz  Don't thank me yet. We need the original commands to actually work, which they don't currently with that DeprecatedCommand object. You'll see that it's overwriting the run command.

danshultz:
Ha, thanks, I'll wire that up now....

womp womp womp on my :)

achambers:
I added a couple of **achambers/ember-cli-deploy** specific commands there
too. Do we need to somehow replicate them, for the time being, for people
that might have upgraded from my repo to this one?

I am currently adding a post install warning to users installing v 0.4.0+
that look to have the config from my repo set ( it sure how else to tell if
they are upgrading from my repo). But I'm not sure that's enough. We don't
really any to break their experience either.

danshultz:
@achambers I'm using your readme still here --> https://www.npmjs.com/package/ember-cli-deploy although if you have your old repo, maybe put it up as ember-cli-deploy-old or something so I can have that as reference.

At the moment, I'm cross checking your readme to update the commands from the original ember-cli-deploy (deploy activate and deploy:versions). I'll have a PR momentarily with some added comments for you to review

achambers:
I think I renamed it to ember-cli-deploy-original.

I'm about to take lunch shortly so can chat about things on slack if you
want.

lib/commands/deprecated/activate.js
```diff
@@ -0,0 +1,31 @@
...
+  description: 'Activate a deployed version of index.html',
+  works: 'insideProject',
+
+  anonymousOptions: [
+    '<key>'
+  ],
+
+  availableOptions: [
+    { name: 'environment', type: String, default: 'development' },
+    { name: 'deploy-config-file', type: String, default: 'config/deploy.js' }
```
danshultz: @achambers - I used the new location of config files, I am leaving the notification to users of the original `ember-cli-deploy` package to a later PR.

danshultz:
I should be around we can chat in slack - couple of questions this PR opens up...

1. Users who update will need to install the `ember-deploy-s3` plugin as it looks to be that deploying to s3 was included in the original `ember-cli-deploy` package.
1. Users will most likely have https://github.com/achambers/ember-cli-deploy-redis-index-adapter installed. We'll need to verify that it still works with 0.4.0 if we are aiming not to break them.

I figure we might handle that in the upgrade notes? If we are however breaking those users, should we leave the `ember deploy:versions` and `ember activate` commands as "unsupported"?

achambers:
Yo @danshultz .  Where does this PR currently sit? Is this good to go for `0.4.0`?

danshultz:
This PR is working in its entirety.

This PR re-enables 'deploy: index', 'deploy: assets'  and adds support for 'deploy: versions' and 'activate' which provides command compatibility with previous versions of ember deploy and ember cli deploy.

@lukemelia mentioned removing the deprecation of the deploy: index and deploy: asset command but thinking it through, we did want to remove these commands in favor of using 'deploy' which does the build and deploying all assets and the index. 

I'm out at the moment but if others can chime in on their opinions I can make any changes to this PR today. Otherwise I think this is ok to merge unless anyone thinks we should do something different.

lukemelia:
My primary concern here is that I don't think `deploy:activate` should be deprecated, and certainly not before we have another way for someone to deploy without immediately activating the new version. @achambers @danshultz Agree/disagree?

danshultz:
Ah, I see the confusion. deploy: activate is not deprecated but the bare 'ember activate' is. This was a command from 0.6.0

lukemelia:
Ah, got it. In that case, +1 to merge.

PR #81, opened 2015-03-05
More name fixes

knownasilya:
ember-deploy to ember-cli-deploy.

Kept ember-deploy-addon, since I wasn't sure if that could be changed.


lukemelia:
Thanks!

Issue #70, opened 2015-03-02
npm postInstall message directed at users of achambers/ember-cli-deploy


achambers:
Thought that maybe it might be a good idea to add a post install message for this version that basically says "Things have changed.  If you are upgrading from achambers/ember-cli-deploy then follow the migration instructions at <some link>".

achambers:
Closed by #85

PR #85, opened 2015-03-06
npm post install message for users of v0.0.6

### LINKED ISSUE #69: Upgrade instructions for users of achambers/ember-cli-deploy
- [ ]  Check for signs that the user might have been using the former ember-cli-deploy and warn them. (You can get a good idea by checking to see where the config is stored.)

achambers:
This PR:

- [x] adds a `MIGRATION_STEPS.md` file containing migration instructions for users of `v0.0.6`.
- [x] adds an npm `postinstall` script that warns users upgrading from `v0.0.6` and points them to the migration instructions
- [x] adds a utility function to detect if the user has defined config in the old location used by `v0.0.6`

Closes #69 and #70

### DIFF
--- a/MIGRATION_STEPS.md
+++ b/MIGRATION_STEPS.md
@@ -1,0 +1,116 @@
+##Overview
+
+`achambers/ember-cli-deploy`, `LevelBossMike/ember-deploy` and `tedconf/ember-cli-front-end-builds` are coming together to create the offical Ember CLI deployment tool, `ember-cli/ember-cli-deploy`.
+
+Users upgrading from the `ember-cli-deploy` npm package `<= v0.0.6` to `>= v0.4.0` will need to follow the migration steps below as the core codebase of the package will be changing completely.
+
+While we are trying our best to maintain backwards compatability for users of `<= v0.0.6`, this will only be temporary and users are strongly urged to migrate ASAP.
+
+## Migrate config
+
+Migrate your `<= v0.0.6` config from this:
+
+```javascript
+// config/deploy/staging.js
+
+module.exports = {
+  assets: {
+    accessKeyId: process.env.AWS_ACCESS_KEY,
+    secretAccessKey: process.env.AWS_SECRET,
+    bucket: 'staging-bucket',
+    region: 'eu-west-1'
+  },
+
+  index: {
+    host: 'staging-redis.example.com',
+    port: '1234'
+  }
+};
+```
+
+```javascript
+// config/deploy/production.js
+
+module.exports = {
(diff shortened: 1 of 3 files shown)


lib/utilities/detect-deprecated-config.js
```diff
@@ -0,0 +1,8 @@
+var path = require('path');
+var glob = require('glob');
+
+module.exports = function() {
+  var deprecatedConfigPath = path.join(__dirname, '../../../..', 'config', 'deploy', '*.js');
```
danshultz: Not sure if there is a better way to get to the user's home directory or not but also, shouldn't this be in the following format to handle different platform path spacing (not sure if that matters in our current scenario or not)

```
path.join(__dirname, '..', '..', '..', '..', 'config', 'deploy', '*.js');
```

achambers: Yeh, I'm not sure if there is a better way either mate.  That's actually the main reason I put it out there to you guys.

Will make the change to the `path.join`.

jasonkarns: I'm fairly certain that `path.join` will normalize the '/' for you, so there's no need to artificially break the segments up as separate arguments. And indeed, I confirmed this is so (at least on Win11/Node 0.10 and 0.12)

screen shot 2015-03-06 at 9 31 24 am

MIGRATION_STEPS.md
```diff
@@ -0,0 +1,111 @@
...
+    },
+    assets: {
+      accessKeyId: 'prod-access-key',
+      secretAccessKey: 'prod-secret',
+      bucket: 'prod-bucket'
+      region: 'eu-west-1'
+    }
+  }
+};
+```
```
lukemelia: Perhaps throw in a line about using environment variables instead of committing secrets.

achambers: Oh yeh. Will do

MIGRATION_STEPS.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Serving of index.html
+
+## Unsupported commands
+
+A number of commands became deprecated in `v0.4.0` and will become unsupported in future versions very soon.
+
+- instead of `ember deploy:index` and `ember deploy:assets`, please use `ember deploy`
+- instead of `ember activate`, please use `ember deploy:activate`
+- instead of `ember deploy:versions`, pleae use `ember:list`
```
lukemelia: s/plea/please/ and s/ember:list/ember deploy:list/

Issue #47, opened 2015-02-27
Deployment pipeline for 0.5.0


ryanto:
# Deployment pipeline

We can look at deploys are a series of steps that change the state of our production application. Our plan is to provide hooks that represent the common steps that you would find in the deployment lifecycle of an Ember application. Deploy adapters will be able to hook onto one or more of the pipeline steps to add functionality to the deploy.

### Proposed steps

1. willDeploy
1. build
1. update
1. activate
1. didDeploy

Questions / comments:
* Is it worth providing an API for adding a step?

### Hooks

Adapters will register themselves with one or more of the hooks to provide functionality. Registration is done by having a function named `deployHookName` on the adapter object.

```javascript
S3Adapter = {
  deployAssets: function(deploy) {
    var results = this.uploadAssets();
    ...
  }
}
```
*Edited to match luke's comment below.*

Questions / comments

* Does order matter? If I register two adapters to ``build`` hook is there a situation where I need to control which gets run first?

### Default adapters

By default ember-cli-deploy will ship with a `BuildDistAdapter` that hooks into the `build` step. This runs `ember build`. It's a good example of a simple adapter + hook, 99% of people will use it, and it easy to override if you need different build-to-dist functionality.

Questions / comments:

* Does this mean we need to provide an API for removing adapters from hooks?

### Shared state and deploy metadata

Adapters may want to share data with other adapters that may run later in the lifecycle. For example a post-deploy adapter may want to summarize all the work done by your build, asset, and index adapters.

There will be an API to read the return values from other adapters. Consider it unstable and not encouraged, but it will be there to see if this is something useful to provide. If we notice of pattern of adapters sharing information we can hopefully provide a shiny API.

lukemelia:
Perhaps an adapter should simply implement the methods that it wants to use (a la ember-cli's addon API).

```javascript
S3Adapter = {
  deployAssets: function(deploy) {
    var results = this.uploadAssets();
    ...
  }
}
```

Also, adapters should likely be returning promises from the hooks since most operations will be async.

ryanto:
Looks good.

:+1: on adapters returning promises. I'll open a new ticket for adapter API.

achambers:
:+1:  Looks good.

ryanto:
@achambers Yup

achambers:
[...]

PR #88, opened 2015-03-07
Plugin architecture documentation

ryanto:
This tries to outline how plugins should look and function. I modeled this off what my dream API would look like for building something like front_end_builds as a plugin. 

Tear this apart!

### DIFF
--- a/rfc/plugins.md
+++ b/rfc/plugins.md
@@ -117,28 +117,42 @@ module.exports = {
 
 ### Hooks
 
+These hooks (part of a typical deployment process) are available for plugins to implement:
+
 ```
-    willDeploy: --> runs before anything happens. good opportunity for plugins to validate
-                    configuration or other preconditions
+willDeploy: --> runs before anything happens. good opportunity for plugins to validate
+                configuration or other preconditions
 
-               /-- willBuild    confirm environment
-              /
-    build --------> builds app assets, documentation, etc.
-              \
-               \-- didBuild     manipulate index.html, validate assets
+           /-- willBuild    confirm environment
+          /
+build --------> builds app assets, documentation, etc.
+          \
+           \-- didBuild     manipulate index.html, validate assets
 
-               /-- willUpload   confirm remote servers(S3, Redis, Azure, etc.)
-              /
-    upload -------> puts the assets somewhere(S3, Redis, Azure, Rackspace, etc.)
-              \
-               \-- didUpload    notify APIs (slack, pusher, etc.), warm cache
+           /-- willUpload   confirm remote servers(S3, Redis, Azure, etc.)
+          /
+upload -------> puts the assets somewhere(S3, Redis, Azure, Rackspace, etc.)
+          \
+           \-- didUpload    notify APIs (slack, pusher, etc.), warm cache
 
-               /-- willActivate   create backup of assets, notify APIs, uninstall earlier versions
-              /
-    activate -------> make a new version live (clear cache, swap Redis values, etc.)
-              \
-               \-- didActivate    notify APIs, warm cache
+           /-- willActivate   create backup of assets, notify APIs, uninstall earlier versions
+          /
+activate -------> make a new version live (clear cache, swap Redis values, etc.)
+          \
+           \-- didActivate    notify APIs, warm cache
 
-    didDeploy: --> runs at the end of a full deployment operation.
+didDeploy: --> runs at the end of a full deployment operation.
 ```
 
+In addition, there are a few more specialized hooks that plugins may implement:
+
+```
+discoverVersions: --> should return a promise resolving to an array of version objects. Each
+                      version object _must_ have an `id` property. Each version _may_ have one
+                      or more of the following properties:
+
+                      `timestamp`:   (Date) when the version was created
+                      `revision`:    (String) reference of version in SCM
+                      `creator`:     (String) email address of developer who deployed the version
+                      `description`: (String) summary of the version
+```
--- a/rfc/plugins.md
+++ b/rfc/plugins.md
@@ -113,6 +113,7 @@ module.exports = {
     }
   }
 };
+```
 
 ### Hooks
 
--- a/rfc/plugins.md
+++ b/rfc/plugins.md
@@ -37,8 +37,9 @@ Plugins are ember-cli addons. They will also have the keyword
 ```
 
 By default, any plugin that is installed will be loaded, similar to
-ember-cli addons. In future versions, we may add a configuration option
-to disable this automatic use of plugins, to allow more flexibility.
+ember-cli addons. The plugin order is determined by how ember-cli orders
+addons (based on `before:`/`after:` properties). To override this, see
+Advanced Plugin Configuration and Ordering below.
 
 ### Plugins are provided by ember-cli addons
 
@@ -49,8 +50,9 @@ ember-cli-deploy. For example:
 ```javascript
 module.exports = {
   name: 'ember-cli-deploy-example-plugin',
-  createDeployPlugin: function(){
+  createDeployPlugin: function(options){
     return {
+      name: options.name,
       willDeploy: function:(deployment){
         // do something during the willDeploy phase of the pipeline
       },
@@ -70,6 +72,10 @@ creation of the plugin object to make any contextual decisions necessary.
 Finally, it is agnostic with respect to the type of object the plugin is.
 It may be a POJO, a subclass of CoreObject, or maybe an ES6 class.
 
+The `options` argument passed to `createDeployPlugin` will have a `name`
+property. Usually, the `name` will be the plugin name sans the `ember-cli-deploy-`
+prefix, unless a name has been specified as described in Advanced Plugin
+Configuration below.
 
 ### The `deployment` object
 
@@ -163,4 +169,32 @@ discoverVersions: --> should return a promise resolving to an array of version o
                       `revision`:    (String) reference of version in SCM
                       `creator`:     (String) email address of developer who deployed the version
                       `description`: (String) summary of the version
+
 ```
+
+### Advanced Plugin Configuration
+
+As mentioned above, by default, all plugins from installed addons will be loaded, and
+ordered based on ember-cli's order of the addons. Developers may have advanced use cases
+for specifying the order of plugins, disabling plugins, or configuring a single plugin to
+be configured and used twice.
+
+If you want to opt-into this configuration, you can set the `plugins` property in your `config/deploy.js` file at either the top-level (for global configuration), or under an environment (for per-environment configuration).
+
+```
+plugins: ["s3-assets", "s3-index", "notify-slack"]
+```
+
+Any plugins not included in the list will not have their hooks executed.
+
+To include a plugin twice, alias it using a colon.
+
+```
+plugins: ["s3-assets:foo-assets", "s3-assets:bar-assets", "s3-index", "notify-slack"]
+```
+
+The name specified after the colon will be passed as the `name` property
+of the `options` argument to the addon's `createDeployPlugin` method. Plugins
+should use their name to retrieve configuration values. In this example,
+the foo-assets instance of the s3-assets plugin could have different configuration
+than the bar-assets instance does.
--- a/rfc/plugins.md
+++ b/rfc/plugins.md
@@ -117,27 +117,27 @@ module.exports = {
 ### Hooks
 
 ```
-willDeploy: --> runs before anything happens. good opportunity for plugins to validate
-                configuration or other preconditions
+    willDeploy: --> runs before anything happens. good opportunity for plugins to validate
+                    configuration or other preconditions
 
-           /-- willBuild    confirm environment
-          /
-build --------> builds app assets, documentation, etc.
-          \
-           \-- didBuild     manipulate index.html, validate assets
+               /-- willBuild    confirm environment
+              /
+    build --------> builds app assets, documentation, etc.
+              \
+               \-- didBuild     manipulate index.html, validate assets
 
-           /-- willUpload   confirm remote servers(S3, Redis, Azure, etc.)
-          /
-upload -------> puts the assets somewhere(S3, Redis, Azure, Rackspace, etc.)
-          \
-           \-- didUpload    notify APIs (slack, pusher, etc.), warm cache
+               /-- willUpload   confirm remote servers(S3, Redis, Azure, etc.)
+              /
+    upload -------> puts the assets somewhere(S3, Redis, Azure, Rackspace, etc.)
+              \
+               \-- didUpload    notify APIs (slack, pusher, etc.), warm cache
 
-           /-- willActivate   create backup of assets, notify APIs, uninstall earlier versions
-          /
-activate -------> make a new version live (clear cache, swap Redis values, etc.)
-          \
-           \-- didActivate    notify APIs, warm cache
+               /-- willActivate   create backup of assets, notify APIs, uninstall earlier versions
+              /
+    activate -------> make a new version live (clear cache, swap Redis values, etc.)
+              \
+               \-- didActivate    notify APIs, warm cache
 
-didDeploy: --> runs at the end of a full deployment operation.
+    didDeploy: --> runs at the end of a full deployment operation.
 ```
 
--- a/rfc/plugins.md
+++ b/rfc/plugins.md
@@ -1,0 +1,233 @@
+# Plugins
+
+A plugin is an ember-cli addon that hooks into the ember-cli-deploy
+pipeline in order to add functionality to a deploy. Example addon
+functionality would be uploading assets to S3, writing the index.html to
+Redis, or notifying a Slack channel a deploy has completed.
+
+Plugins should focus on doing a specific task. Ember developers will
+compose and combine multiple plugins to fine tune their deployment
+process.
+
+### Identifying plugins
+
+Plugins are published to npm as ember-cli addons. They will also have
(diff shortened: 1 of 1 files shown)


rfc/plugins.md
```diff
@@ -0,0 +1,235 @@
+# Plugins
+
+A plugin is an ember-cli addon that hooks into the ember-cli-deploy
+pipeline in order to add functionality to a deploy. Example addon
+functionality would be uploading assets to S3, writing the index.html to
+Redis, or notifying a Slack channel a deploy has completed.
+
+Plugins should focus on doing a specific task. Ember developers will
+compose and combine multiple plugins to fine tune their deployment
+process.
```
lukemelia: Most plugins designed for distribution should focus on a specific task. However, teams wishing to fully customize their deployment process via a private or in-repo addon need not limit the addon to a specific task.

rfc/plugins.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Identifying plugins
+
+Plugins are published to npm as ember-cli addons. They will also have
+the keyword ``ember-cli-deploy-plugin``.
+
```
lukemelia: Let's add example snippet from a package.json here.

rfc/plugins.md
```diff
@@ -0,0 +1,235 @@
...
+Any plugin that is installed will be loaded by default, similar to
+ember-cli addons.
+
+### Base plugin object
+
+Plugins will extend a base object provided by ember-cli-deploy. The API
```
lukemelia: This means that all ember-cli-deploy-plugins will need a dependency on ember-cli-deploy. Is that weird?

rfc/plugins.md
```diff
@@ -0,0 +1,235 @@
...
+  ...
+});
+```
+
+### Hooks
+
+Plugins hook into the deployment pipeline with function names that match
+the steps in the pipeline process. These hooks will scoped under a
+special hooks object inside the plugin so that any private or plugin
+functions do not clash with any hook names that we add in
```
lukemelia: The `hooks` object is going to make method context weird. How about of instead the properties names for the functions are prefixed with `deploy:`. e.g.

```javascript
var S3Plugin = DeployPlugin.extend({
  "deploy:updateAssets": function(...){
    ...
  },
  "deploy:activate": function(...){
    ...
  },
});
```

rfc/plugins.md
```diff
@@ -0,0 +1,235 @@
...
+This task would run as `ember deploy s3-plugin:doctor`.
+
+Is this worth providing or should we just let plugin authors do this
+themselves in their ember-cli-addons?
+
+I think it might be worth exploring because there might be many common
+tasks that these plugins provide. We could have an `ember deploy:doctor`
+that runs the doctor task of each plugin and reports back their return
+values. Things like `dry-run`, `doctor`, or `setup` might be common
+tasks you would find across many different plugins.
```
lukemelia: I say let's leave commands out of our API and RFC. Plugins can certainly provide them if they want to. If were to implement additional commands, though, we would delegate to hook methods of plugins not to commands.

rfc/plugins.md
```diff
@@ -0,0 +1,235 @@
...
+      awsSecretAccessKey: process.env.AWS_SECRET_ACCESS_KEY
+    }
+  }
+};
+```
+
+This data will then be available to the `S3Plugin` function's as
+`this.pluginConfig`.
+
+
```
lukemelia: Can we add a section to this document specifying the hook methods? We can incorporate the results of the conversation from #47

rfc/plugins.md
ryanto: Yeah agreed. Will remove this.

Twinkletoes:
I don't think plugins should be singletons. What if I need to deploy to two locations (e.g. my B2B and B2C sites)? I'd want two plugins of the same type somewhere in my deploy chain.

lukemelia:
@Twinkletoes currently we're thinking of deploying to different locations (as your B2B and B2C example) as different deployment targets, each with their own config that would be targeted by separate commands. e.g. `ember deploy --deploy-target=b2b-prod` and `ember deploy --deploy-target=b2c-prod`

Twinkletoes:
@lukemelia What if most of the deploy stack is shared but there are two frontends (e.g. two DNS names or two cloudfronts, but same asset store)

lukemelia:
@Twinkletoes in that case the assets only need to be deployed once. You would not typically be making any changes to your DNS or CDN as part of a deployment would you?

Twinkletoes:
The deployment script should deploy. At the end of it the site should be up and working. If that involves creating missing pieces then so be it.

In my case we deploy repo branches as-live, using subfolders in the bucket, one per branch. The first time a branch gets deployed, we need to create cloudfront distributions and DNS entries

lukemelia:
@Twinkletoes I definitely think that we should support your use case with ember-cli-deploy. I don't think that the creation of cloudfront distributions and DNS entries will be part of standard OSS plugins (though I could be wrong). I think the important thing to focus on for that is will the deployment pipeline have the right hooks that you can combine open source plugins with a custom one that performs distribution and DNS setup as-needed. Perhaps the `willUpload` or `didUpload` hook would be suitable?

Twinkletoes:
I was expecting to create an open source plugin to do the cloudfront stuff, but that's a specific use case that the deploy pipeline should be flexible enough to manage.

I don't see much difference between the redis activate and the cloudfront create distribution. Neither are about deploying files - they're both in the post-install step

My point is more that there may be more than one post install step, run-ordering matters, and the same plugin might be used more than once in the process. .

rfc/plugins.md
achambers: I like the idea of having a base Plugin that is extended.  Similar to extending an `ember-cli` Task.  But now you mention it @lukemelia, maybe it is weird to have a dependency on `ember-cli-deploy`.  But then again, it is a plugin for `ember-cli-deploy`, so maybe it isn't that weird.  This comment added zero value whatsoever ;)

rfc/plugins.md
```diff
@@ -0,0 +1,235 @@
...
+  deployConfig: { ... },
+  pluginConfig: { ... }
+}
+```
+
+Property | file | info
+--- | --- | ---
+`ui` | - | The ember-cli UI object that can be used to write to stdout.
+`appConfig` | `config/environment.js` | The application configuration.
+`deployConfig` | `config/deploy.js` | The deploy configuration file.
```
achambers: Do we need `deployConfig` if we are also passing `pluginConfig` in?  Should a plugin care about any config other than it's own?  If the plugins are supposed to be independent entities then I'd imagine everything they need should be in their own config, no? What's an example of a use case where a plugin may need access to the rest of the stuff in `config/deploy.js`?

rfc/plugins.md
```diff
@@ -0,0 +1,235 @@
...
+Hook functions can return a promise to block the deployment pipeline.
+Since most deployment involves some sort of IO it makes senses that most
+plugins will want an async function to complete before continuing to the
+next step. If a plugin does not return a promise the next pipeline step
+begins running.
+
+Additionally, plugins can return a POJO, directly or from a promise,
+with data about what happened. Future plugins can access that data,
+more on that later.
+
```
achambers: Does it make sense to make a plugin HAVE to return a Promise?  Seems like it might simplify things for us and for the people writing plugins if there is only one thing they can return.  We could use something similar to `ember-cli` Tasks where a plugin can either:

- resolve with undefined or a POJO
- reject with an Error instance if the error is unhandled or;
- reject with undefined if it was handled.

lukemelia: We can call `rsvp.resolve(returnValue)` on the return value of each hook, which will nicely handle non-promises and promises alike.

rfc/plugins.md
```diff
@@ -0,0 +1,206 @@
...
+  "version": "0.4.0",
+  // ...
+  "devDependencies": {
+    "ember-cli": "0.2.0",
+    "ember-cli-deploy": "0.4.1"
+    // ...
+  },
+  "keywords": [
+    "ember-addon",
+    "ember-cli-deploy-plugin"
```
achambers: Something to think about here, and I don't believe it's mentioned anywhere else in this document, but this keyword isn't really relevant other than to tell indicate to people (or machines) looking at `package.json` that this is an ember-cli-deploy addon.

Our code actually won't see this keyword as, by the time we get access to the addons via `this.project.addons`, we don't have access to the package.json (this is my understanding anyway).

It is for this reason that we have previously set a `type` property on the plugin object and the `adapter-registry` interrogates this property ([here](https://github.com/ember-cli/ember-cli-deploy/blob/master/lib/utilities/adapter-registry.js#L36)) to see whether the addon is an ember-cli-deploy plugin.

Maybe we should mention this in the document.

achambers: Thinking about it there, we can just check for the presence of the createDeployPlugin function. Ignore me. Carry on.

lukemelia: @achambers I'm pretty sure it is accessible at `addon.pkg.keyworks`. That said, your suggestion of looking for that method seems reasonable, too. We don't get easy npm filtering without the keyword though...

ryanto:
:+1: thanks luke

rfc/plugins.md
```diff
@@ -0,0 +1,206 @@
...
+                      or more of the following properties:
+
+                      `timestamp`:   (Date) when the version was created
+                      `revision`:    (String) reference of version in SCM
+                      `creator`:     (String) email address of developer who deployed the version
+                      `description`: (String) summary of the version
+
+```
+
+### Advanced Plugin Configuration
```
danshultz: I really like this section for the advanced plugin configuration stuff! I think it really addresses the use cases that have come up lately

:+1:

samselikoff:
This is looking really great. Love the activate/discoverVersions hooks. I think this is us saying, "here's some structure for you based on what the community has discovered, but we're not going to prescribe too much."

samselikoff:
What do you all think about the host app also being able to write code in the various hooks? Essentially, we should give them the ability to modify their deploy pipeline without having to make a plugin.

```js
//config/deployPipeline.js

export default {
  upload: function(deploy, config) {
    this.run('s3:upload'); // runs the `ember-cli-deploy-s3 plugin's upload hook
    // do some of my own stuff
  }
};
```

lukemelia:
@samselikoff A similar idea was brought up by @danshultz. The Advanced Plugin Configuration section at the bottom is meant to address at least a portion of what you're describing without additional Javascript API surface area. I guess my vote would be to wait to see if we need it, but if you think this is really important, I'm open to adding it.

samselikoff:
right. I guess the question is, will people's deploy pipelines be standard enough that they could get away with constructing them using available plugins only? If devs are creating bespoke plugins just to "wrap" or "glue together" other plugins for their particular pipeline, I think we've made a mistake.

I'm okay waiting on this though.

rfc/plugins.md
```diff
@@ -0,0 +1,206 @@
...
+module.exports = {
+  production: {
+    "example": {
+      bucket: "my-app-production",
+      awsAccessKeyId: process.env.AWS_ACCESS_KEY_ID,
+      awsSecretAccessKey: process.env.AWS_SECRET_ACCESS_KEY
+    }
+  }
+};
+```
```
achambers: Lads, what are your thoughts on structuring the config like the ember-cli-config?  ie:

```javascript
  var ENV = {
    plugins: ["s3-assets", "s3-index", "notify-slack"]
  };

  if (environment === 'development') {
    ENV.redis = {
      host: 'localhost',
      port: 1234
    };
  }

  return ENV;
}
```

The reason I ask is as I was having a crack at implementing this RFC actually code, I figured it'd be nice to use the ConfigurationReader object for both the `config` and `appConfig` properties that are available on the `deployment` object.  By making the config be of the same format as the ember-cli config object, we could have a generic ConfigurationReader object and Config model that doesn't care which config it is responsible for.

In making this change, we could provide a blueprint that will migrate your existing config object to the new format if need be.

While this is not essential, it feels like maybe it might be a good idea to carry on the conventions ember-cli is using instead of creating our own.

Thoughts?

lukemelia: In theory, I think aligning to ember-cli conventions is a good idea. I think we probably want the top level nodes of the config object to be the deployment targets.

Also, I see a future in which we are able to move our config into ember-cli's config infrastructure, once it supports "non-runtime" configuration. Something like this: https://gist.github.com/rwjblue/ecb483b15da77582d0b9

We shouldn't wait on that, though.

achambers: > I think we probably want the top level nodes of the config object to be the deployment targets.

@lukemelia: Just out of curiosity, what's the reasoning behind this?

Based on the advanced config below, this potentially won't strictly be the case either right?  As `plugins` can be specified at the top level.

Twinkletoes: I think the config is upside down. I was talking to Aaron about my ideas at
Ember London and I'm writing a new RFC.


Essentially I think a plugin should do exactly one thing:
- Deploy some files to s3
- Deploy a key-value to redis
- Create an s3 bucket
- Send a hip chat message
Etc

The developer can then set up a config that is an ordered list of tasks. A
plugin that uploads files should take a glob. It shouldn't know which files
it should upload in advance (eg assets). For example if I have a b2b and
b2c version, my images and CSS might differ so I might have two asset
upload tasks with different targets.

Third party plugins are now much easier to integrate - they just get added
to the list of tasks and resolved by name.

If task x fails, tasks 1->x-1 are rolled back. It's up to a plugin
how/whether  it implements the rollback function.

The only unknown here is how to bind input of task x to output of task x-1.
I'd like to support js expressions here.

For ease of setting up a lukemelia deploy, we would have a config blueprint
which sets up the task list for that deploy process. We can have other
blueprints inc third party ones for other common processes.

Thoughts?

Bryan

samselikoff: when you say "ordered list of tasks" are you proposing getting rid of the hooks and having each host app explicitly specify its own workflow, a la grunt? We thought of this but decided against it, because we want ember-cli-deploy to provide more structure than that. Plugins should execute on certain hooks by default, with the ability to override. Overriding should be a minority case, and would be handled in the config as specified below.

or am I misunderstanding your point?

Twinkletoes: You're understanding me but disagreeing with me.

If you go down the current route, then it's very hard to bend the workflow
to needs other than the default usage. For example my desired workflow
deploys everything to S3 and then configures cloudfront and Route 53, then
sends a hipchat message. That's very hard to implement with the current
design of ember-cli-deploy.

In some of my workflows I deploy to two different S3 buckets. If plugins
are singletons then this type of deployment is impossible.

If instead the workflow is setup in a task list, then the default task list
can still give the ease of use of a blueprint:
`ember generate deploy:config lukemelia`

It's just that the workflow is declarative rather than compiled into an
addon. If I want a different workflow for my project, I edit a config
rather than having to write a new addon. I can even take the blueprinted
config and add a notification plugin to the end of it. It makes modifying
the default workflow trivial.

It also makes the deploy system much easier to build, and much easier to
build new addons for. Each plugin now has one hook, "run", which takes a
config hash and returns a hash which can be bound into the input hash of
future tasks.

In conclusion I'd like to see a flexible deploy addon, with the lukemelia
deploy process as a special case, probably as the default configuration.
But multiple other special cases should also be possible with the same
addon, or with new third party deploy plugins (e.g. cloudfront, hipchat
etc).

On 26 March 2015 at 12:12, Sam Selikoff <notifications@github.com> wrote:

> In rfc/plugins.md
> <https://github.com/ember-cli/ember-cli-deploy/pull/88#discussion_r27207238>
> :
>
> > +By convention, plugin configuration should be kept in `config/deploy.js` and scoped by
> > +the plugin's name. e.g. for an `ember-cli-deploy-example` plugin, the configuration would
> > +go in:
> > +
> > +```javascript
> > +// config/deploy.js
> > +module.exports = {
> > +  production: {
> > +    "example": {
> > +      bucket: "my-app-production",
> > +      awsAccessKeyId: process.env.AWS_ACCESS_KEY_ID,
> > +      awsSecretAccessKey: process.env.AWS_SECRET_ACCESS_KEY
> > +    }
> > +  }
> > +};
> > +```
>
> when you say "ordered list of tasks" are you proposing getting rid of the
> hooks and having each host app explicitly specify its own workflow, a la
> grunt? We thought of this but decided against it, because we want
> ember-cli-deploy to provide more structure than that. Plugins should
> execute on certain hooks by default, with the ability to override.
> Overriding should be a minority case, and would be handled in the config as
> specified below.
>
> or am I misunderstanding your point?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ember-cli/ember-cli-deploy/pull/88/files#r27207238>.
>



-- 
Bryan Crotaz
Managing Director
Silver Curve

rfc/plugins.md
```diff
@@ -0,0 +1,206 @@
...
+
+### The `deployment` object
+
+For each high-level ember-cli-deploy operation, a `deployment` object is created.
+This object is passed to each hook that is invoked on the plugins. It has a number
+of properties that may be of use to a plugin:
+
+Property | file | info
+--- | --- | ---
+`ui` | - | The ember-cli UI object that can be used to write to stdout.
```
achambers: @lukemelia What are your thoughts about supplying the `analytics` and `project` objects in this `deployment` object as well?  The Ember CLI BuildTask  requires both of them I believe.  The `project` object is present on the `config` property however this doesn't quite feel right to me.

Thoughts?

lukemelia:
Merged into `0.5.0-dev` branch at b4797999bfbbb191841efdf7b8c09b8d398db79b

PR #100, opened 2015-03-13
Update config around production-like environments

Soliah:
I understand stuff is changing real soon with #59 but I was struggling with getting staging setup to behave exactly like production in terms of fingerprinting, minifying etc right now.

There are a number of other options that are required in ember-cli if a build exactly the same as `--environment production` is desired, which i've now added to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,12 +150,26 @@ var isProductionLikeBuild = ['production', 'staging'].indexOf(env) > -1;
 
 var app = new EmberApp({
   fingerprint: {
+    enabled: isProductionLikeBuild,
     prepend: 'https://subdomain.cloudfront.net/'
   },
+  sourcemaps: {
+    enabled: !isProductionLikeBuild,
+  },
+  minifyCSS: { enabled: isProductionLikeBuild },
+  minifyJS: { enabled: isProductionLikeBuild },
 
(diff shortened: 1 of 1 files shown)


LevelbossMike:
Thanks for improving the README! Unfortunately this is still a pain to do in ember-cli and also other addons are a problem with this approach because they can't be forced to add productionLike assets via the app's Brocfile (ember-faker is an example). Bloats the build but does not change the app's behaviour much of course. 

Would be great if you could also add the vendorFiles-property which is needed to add the production like builds for handlebars and ember itself for `vendor.js`:

```
  vendorFiles: {
    'handlebars.js': {
      staging:  'bower_components/handlebars/handlebars.runtime.js'
    },
    'ember.js': {
      staging:  'bower_components/ember/ember.prod.js'
    }
```

Soliah:
Think there's an bunch of issues in the ember-cli repo around multiple environments so hopefully there will be a simpler solution soon.

Just pushed up the `vendorFiles` addition.

Soliah:
Actually, is adding vendorFiles necessary? Adding it breaking `ember build` for me.

Soliah:
Ah I don't have handlebars.

LevelbossMike:
Thx again! :+1:

PR #113, opened 2015-03-22
Provide better error support for missing environment config

### LINKED ISSUE #101: Provide Better Error Support for Missing Environment Deploy Config
The default build environment is `development` and if you run an `ember deploy` without specifying the build environment it can provide confusion as most will not have a config for `development`. Additionally, the error for a missing environment section is a nondescript null error. 

When I run a deploy for an environment for which no config section exists
Then an error should inform me that my config section for `<environment>` is missing.

achambers:
This PR will add better messaging when a user tries to run an `ember-cli-deploy` command for an environment for which there exists no configuration.  This most commonly manifests itself when a user does not supply an `--environment` option, running in the default `development` env for which they have not provided config.

This PR explains what the problem is and points the user to the README for more info.

Closes #101

### DIFF
--- a/node-tests/unit/utilities/configuration-reader-test.js
+++ b/node-tests/unit/utilities/configuration-reader-test.js
@@ -90,6 +90,19 @@ describe('ConfigurationReader', function() {
         });
       }).to.throw('Cannot load configuration file \'' + path.join(root, configPath) + '\'. Note that the default location of the ember-cli-deploy config file is now \'config/deploy.js\'');
     });
+
+    it('raises an error if a config doesn\'t exist for the current environment', function() {
+      var configPath = './node-tests/fixtures/config/deploy.js';
+
+      expect(function() {
+        new ConfigurationReader({
+          configFile: configPath,
+          ui: ui,
+          project: project,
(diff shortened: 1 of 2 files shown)


node-tests/unit/utilities/configuration-reader-test.js
```diff
@@ -90,6 +90,23 @@ describe('ConfigurationReader', function() {
...
+
+      try {
+        new ConfigurationReader({
+          configFile: configPath,
+          ui: ui,
+          project: project,
+          environment: 'non-existent-env'
+        });
+
+        expect(false).to.be.ok;
```
achambers: Have never really known what the best practice is here when asserting that an Error is thrown.  Any suggestions would be much appreciated.

lukemelia:
@achambers here's what you are looking for: http://api.qunitjs.com/throws/

achambers:
@lukemelia We're not using qunit in the ember-cli-deploy `node-tests` but chai has the same thing.  Thanks.  Updating now.

achambers:
Changed.  Should be ready to merge if you have a minute @lukemelia

PR #117, opened 2015-03-23
[WIP] Integration test through commonly used plugins

danshultz:
In order to help prevent breaking some of the most used plugins and help identify issues with using commonly used plugins as ember-cli-deploy changes, these tests will serve as details of how configurations work along with passing tests to identify breakages.

Items of note currently in flight or in the future

* Need to support running the "plugin-integration" tests outside of normal `npm test` command as these tests will require quite a bit of env setup
* Currently I opted to simply `npm install` the plugin vs including it in the package.json. Alternatively, I thought about creating a set of package.json files to use with these tests. That does open up quite a few rabbit holes however going down one off package.json files. I tried to limit what is included in the current package.json for ember-deploy-cli because node and dependencies. Also, that would be alot to maintain.
* leveraging existing ember app in test/app for this to work and deploy.

Have quite a bit of cleanup but here's are first pass. I'll most likely need to update the travis.yml file to provide redis access.

### DIFF
--- a/lib/utilities/adapter-registry.js
+++ b/lib/utilities/adapter-registry.js
@@ -46,6 +46,8 @@ module.exports = CoreObject.extend({
     var adaptersToLoadForTest = process.env['LOAD_ADAPTERS'];
     if (!adaptersToLoadForTest) { return; }
 
+    this.adapters['assets']['null-adapter'] = CoreObject.extend({ upload: function () {} });
+
     adaptersToLoadForTest.split(',').forEach(function (adapterName) {
       var root = this.project.root;
       var addonPath = path.join(root, 'node_modules/' + adapterName);
--- a/lib/utilities/adapter-registry.js
+++ b/lib/utilities/adapter-registry.js
@@ -1,13 +1,17 @@
+/* jshint node: true */
 var CoreObject     = require('core-object');
 var UnknownAdapter = require('./assets/unknown');
 var ShaAdapter     = require('./tagging/sha');
 var merge          = require('lodash-node/modern/objects/merge');
+var Addon = require('ember-cli/lib/models/addon');
+var path = require('path');
 
 module.exports = CoreObject.extend({
   init: function() {
     if (!this.project) { return; }
 
     this.project.addons.forEach(this._mergeDeployAdapters.bind(this));
+    this._addAdaptersForTesting();
   },
 
   lookup: function(type, adapterName) {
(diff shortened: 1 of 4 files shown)


lib/utilities/adapter-registry.js
```diff
@@ -1,13 +1,17 @@
+/* jshint node: true */
 var CoreObject     = require('core-object');
 var UnknownAdapter = require('./assets/unknown');
 var ShaAdapter     = require('./tagging/sha');
 var merge          = require('lodash-node/modern/objects/merge');
+var Addon = require('ember-cli/lib/models/addon');
```
danshultz: Need to line these up

lib/utilities/adapter-registry.js
```diff
@@ -36,5 +40,36 @@ module.exports = CoreObject.extend({
     if (addon.type === 'ember-deploy-addon') {
       this.adapters = merge(this.adapters, addon.adapters);
     }
+  },
+
+  _addAdaptersForTesting: function () {
```
danshultz: Support test shim for loading plugins not in the package.json. Downside with this approach is we bypass DAG which may cause some issues in the future with before/after hooks, etc.

package.json
```diff
@@ -35,7 +35,9 @@
     "express": "^4.8.5",
     "github": "0.2.3",
     "mocha": "^2.0.1",
-    "sinon": "^1.12.1"
+    "redis": "^0.12.1",
+    "sinon": "^1.12.1",
+    "then-redis": "^1.3.0"
```
danshultz: added then-redis to the dev dependencies to be able to inspect and query redis

node-tests/plugin-integration/ember-deploy-redis-test.js
```diff
@@ -0,0 +1,40 @@
+/* jshint expr:true */
```
danshultz: quite a bit of cleanup to be done here, but here is the gist thus far and it works locally at the moment.

LevelbossMike:
Perhaps it would be a good idea that these smoke tests would live in their own repository. We then wouldn't need to add dependencies on then-redis etc. in ember-cli-deploy directly.

danshultz:
@LevelbossMike I think that would be a good idea as I would also prefer to avoid adding dependencies to ember-cli-deploy that are unnecessary. It would make it a bit simpler also and easier to create multiple package.json files and provide multiple examples of config.

Possibly a directory structure like below to support the multiple apps, etc. This other repository could build based on a hook from `ember-cli-deploy` master pushes maybe?

```
[dshultz|ember-cli-deploy-integrations]$ tree
.
├── ember-deploy-redis
│   ├── Brocfile.js
│   └── package.json
├── ember-deploy-s3
│   ├── Brocfile.js
│   └── package.json
└── test-app
    ├── Brocfile.js
    ├── README.md
    ├── app
    ├── bower.json
    ├── config
    │   └── environment.js
    ├── package.json
    ├── public
    │   ├── crossdomain.xml
    │   └── robots.txt
    ├── testem.json
    ├── tests
    │   ├── helpers
    │   │   ├── resolver.js
    │   │   └── start-app.js
    │   ├── index.html
    │   ├── test-helper.js
    │   └── unit
    └── vendor
```

danshultz:
One other strategy I was thinking about, was doing something similar to what @kategengler did in her ember-try plugin https://github.com/kategengler/ember-try

The test run would execute `npm install --save-dev` for the necessary dependencies for the tests and then do a `git reset --hard` as part of the test run to reset for reach test. In doing that, I would also include a safety check to ensure that it would not do a `git reset --hard` if the index was dirty.

The full integration tests would need a special flag to run or be run single file directly. The tests would primarily run on Travis but could be run locally if a person would like. The tests would leverage environment variables for s3 and redis config

lukemelia:
Closing in favor of work in progress by @ghedamat and others

PR #118, opened 2015-03-23
Update README.md

maxwerr:
I may be totally misreading this, but it looks like

    index_key ||= redis.get('<your-project-name>:current')
    redis.get(index_key)
would set index_key to be the index file, then attempt to get a key of the index file? Or no? Or... redis things I don't know?

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,7 +187,7 @@ require 'redis'
 
 def bootstrap_index(index_key)
   redis = Redis.new
-  index_key ||= redis.get('<your-project-name>:current')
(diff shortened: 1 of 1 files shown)


achambers:
Hi @maxwerr . The "<projectname>:current" key store the revision number which relates to the index file, not the index file itself. So the code is correct as it stands.

achambers:
I'm going to close this for now. Let me know if you have any other questions.

maxwerr:
Ah I got it, thank you. I just misunderstood the process, current points to the key whose value is the index file. Makes sense.

PR #119, opened 2015-03-27
Stub active, list, createTag UnknownAdapter methods

archit:
While developing an index adapter (more on that later), I was getting an unhelpful message;

```
undefined is not a function
TypeError: undefined is not a function
```

This is was because the AdapterRegistry was defaulting to the UnknownAdapter. However unknown adapter was missing the `list` method, so I had to do some manual debugging before I could find it.

### DIFF
--- a/lib/utilities/assets/unknown.js
+++ b/lib/utilities/assets/unknown.js
@@ -2,10 +2,15 @@ var Adapter     = require('../adapter');
 var Promise     = require('ember-cli/lib/ext/promise');
 var SilentError = require('ember-cli/lib/errors/silent');
 
+var errorMessage = function() {
+  var errorMessage = 'You tried to use an unknown adapter: `' + this.name +
+    '`. Please pass a supported adapter-type.';
+  return Promise.reject(new SilentError(errorMessage));
+}
+
 module.exports = Adapter.extend({
-  upload: function() {
-    var errorMessage = 'You tried to use an unknown adapter: `' + this.name +
-                       '`. Please pass a supported adapter-type.';
-    return Promise.reject(new SilentError(errorMessage));
-  }
+  upload: errorMessage,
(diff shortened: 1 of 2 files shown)


lib/utilities/assets/unknown.js
```diff
@@ -7,5 +7,20 @@ module.exports = Adapter.extend({
     var errorMessage = 'You tried to use an unknown adapter: `' + this.name +
                        '`. Please pass a supported adapter-type.';
     return Promise.reject(new SilentError(errorMessage));
+  },
+  activate: function() {
+    var errorMessage = 'You tried to use an unknown adapter: `' + this.name +
```
achambers: Thanks for your contribution. Do you think you could refactor the duplicated code out into a private function?

Also some tests would be appreciated too. :+1:

archit:
Update per review. ping @achambers

archit:
Weird that GitHub is pulling build `222.1` which fails & not `222.2` that passes.

node-tests/unit/utilities/assets/unknown-test.js
```diff
@@ -12,5 +12,23 @@ describe('UnknownAdapter', function() {
       unknown = new UnknownAdapter();
       expect(unknown.upload()).to.be.rejected;
     });
-  })
+  });
+  describe('#activate', function() {
+    it('rejects to print out an error message', function(){
+      unknown = new UnknownAdapter();
```
LevelbossMike: All these tests are basically the same. We can use a function to create the `expect(unknown.<method>)` calls. Also the initialization of the unknown adapter could be moved into a beforeEach function.

archit:
@LevelbossMike tests are consolidate as requested.

LevelbossMike:
Thanks!

PR #130, opened 2015-04-22
[WIP] v0.5.0 Deploy Pipline

achambers:
Moved to #179


achambers:
Lads, figured I'd just create a PR now as it seems ye are happy with the progress on this....Keen to get discussion going here.

lib/tasks/pipeline.js
```diff
@@ -0,0 +1,100 @@
...
+
+    var root = this.project.root;
+
+    this.deployEnvironment = this.deployEnvironment || 'production';
+
+    var dotEnvFilename = '.env.deploy.' + this.deployEnvironment;
+    var dotEnvFilePath = path.join(root, dotEnvFilename);
+
+    if (existsSync(dotEnvFilePath)) {
+      dotenv.load({
```
achambers: Loading env vars from .env.deploy.<environment> files if they exist.  I attempted to create an ember-cli-deploy-dotenv plugin but this wasn't really possible.  Happy to discuss why if needed.  Seems that this is a pretty standard requirement so I think it makes sense to do it in the PipelineTask.

lib/utilities/pipeline.js
```diff
@@ -0,0 +1,73 @@
...
+
+Pipeline.prototype.register = function(hookName, fn) {
+  var pipelineHooks = this._pipelineHooks;
+
+  if (pipelineHooks[hookName]) {
+    pipelineHooks[hookName].push(fn);
+  }
+};
+
+Pipeline.prototype.execute = function(context) {
```
achambers: This all gets pretty gnarly.  I have been trying to make it uncomplicated but not sure how to.  Still, it's a work in progress so it will change.  Happy to take comments, suggestions, advice.

LevelbossMike: naming the functions in the then and catch parts (instead of using anonymous functions) would make it a little bit easier to read/understand otherwise it doesn't seem to be that hard to get what execute does in imo

achambers: Yup. Will do. Thanks mate.

lib/utilities/pipeline.js
```diff
@@ -0,0 +1,73 @@
...
+
+var Promise = require('rsvp').Promise;
+var _       = require('lodash-node');
+
+var chalk = require('chalk');
+var blue  = chalk.blue;
+var red   = chalk.red;
+var grey  = chalk.grey;
+
+function Pipeline(hookNames, options) {
```
LevelbossMike: Any reason why we are not using core-object here? The rest of ember-cli and ember-cli-deploy is using core-object when doing `classy` stuff

achambers: No reason. Just didn't feel the need. Can switch it over

achambers: Actually, I lie slightly there. My thinking was that this will end up being extracted to a library that has nothing to do with ember or ECD. So, was just writing it without any of the existing dependencies. Obviously we can still use core object. I just didn't feel the need.

LevelbossMike: I think using core object would make it easier for other ember-devs to join in. It's not hard with a prototype but when being used to the ember.object model I think it's easier to other devs when being able to work with a 'normal' object.

achambers: Ok...Will switch it over.

lib/tasks/pipeline.js
blimmer: This seems very reasonable to me.

blimmer:
Really nice work on this @achambers , thanks for getting this out there for feedback.

achambers:
Thanks mate. No probs. let me know how it goes if you start testing it out.
I'm working on it pretty actively now. You can see the plugins I'm working
on at https://github.com/zapnito

lukemelia:
Moved to #179

PR #135, opened 2015-05-19
Default Environment to Staging

CoderCookE:
Hey, 
For security reasons we think it makes more sense for the default
to be staging, to prevent any accidents.  I know it's nice to be able to
just quickly type `ember deploy` and that you still need to activate.
But I think it would be better to make the process more deliberate.

-codercooke

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ By default, `ember-cli-deploy` expects a `deploy.js` file in the `config/` direc
 ```js
 module.exports = {
   development: {
-    buildEnv: 'development', // Override the environment passed to the ember asset build. Defaults to 'production'
+    buildEnv: 'development', // Override the environment passed to the ember asset build. Defaults to 'staging'
     store: {
       type: 'redis', // the default store is 'redis'
       host: 'localhost',
@@ -85,7 +85,6 @@ module.exports = {
   },
 
   staging: {
-    buildEnv: 'staging', // Override the environment passed to the ember asset build. Defaults to 'production'
     store: {
       host: 'staging-redis.example.com',
       port: 6379
@@ -98,6 +97,7 @@ module.exports = {
   },
 
    production: {
+    buildEnv: 'production', // Override the environment passed to the ember asset build. Defaults to 'staging'
     store: {
       host: 'production-redis.example.com',
       port: 6379,
--- a/lib/models/config.js
+++ b/lib/models/config.js
@@ -45,7 +45,7 @@ module.exports = CoreObject.extend({
       'assets.gzip': true,
       'assets.gzipExtensions': ['js', 'css', 'svg'],
       'assets.type': 's3',
-      'buildEnv': 'production',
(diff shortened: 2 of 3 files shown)


blimmer:
Seems like it should be even safer and default to development. Then you really wouldn't step on anyone's toes.

LevelbossMike:
@CoderCookE Thanks for this PR.

I may be misunderstanding something here but I don't get why a production default would be `unsafe`. `ember-cli-deploy` is meant to deploy your application somewhere so you'd likely want to deploy a version of your application that is built with a production ready build (concatenated vendor.js and app.js  asset files, support for cdn urls etc.)

Also there shouldn't be anything super confidential in your deployed ember app which would be unsafe to deploy as your users will eventually see the production build anyway...?!

The buildEnv is just meant as a way to be able to deploy production like builds to environments like `staging` where you want the app to behave the same as a production build but maybe use other environment settings for api-hosts etc.

@blimmer I don't see much sense in deploying an development build anywhere because the app wouldn't behave like a production build at all (assets wouldn't be concatenated etc.) Could you elaborate in which use-case you wanted to deploy the development build anywhere?

It's totally possible that I am on the wrong track here but I really don't see much advantage in changing this default option to the staging build. On top it's also pretty annoying to setup a staging build so I don't think would make a good default.

blimmer:
I don't have a direct use for deploying to development environment, I was just thinking that you should always deploy somewhere that would never have adverse side effects. I guess the activation requirement might make this a non-issue, but I agree that ember-deploy deploying to your production environment by default is a bit scary.

LevelbossMike:
Aaah now I see! Just for clarification. That's a build option. This will only affect the build of your app. The default environment your would deploy to is `development`.

If you wanted to deploy to production you'd need to pass the `--environment` flag to the `ember deploy` command because `development` is the default environment that gets passed in the deploy command:

https://github.com/ember-cli/ember-cli-deploy/blob/master/lib/commands/deploy.js#L10

blimmer:
OK, now I understand too. I wonder if @CoderCookE had this same misunderstanding.

LevelbossMike:
I am closing this for now as this seems to be a misunderstanding in how the ember deploy command works. please feel free to reopen if we have missed something here @CoderCookE

CoderCookE:
Ahh ok thanks @LevelbossMike and @blimmer it was definitely some misunderstanding about it how it works.  I appreciate you taking the time to respond.

On a related note we do `deploy` to `development` on occasion (pretty much once per dev machine) which writes to a local redis so that the rails server pulls in the assets locally prepending the assets with 'http://localhost:8020/assets/'

LevelbossMike:
@CoderCookE You should also be able to deploy locally when using your local redis as an index store and s3 (or similar) as asset-store. The advantage in this approach would be to be able to see if assets are correctly cached. You could also use different cdn-urls for development and other environments that way. You'd have to change the fingerprint options though.

PR #141, opened 2015-05-27
Add configuration option to exclude asset files from being deployed.

### LINKED ISSUE #138: Uploading Hidden Files
I updated ember-cli today to 0.2.5. Upon deploying to a staging box, I noticed the following:

```
Built project successfully. Stored in "dist/".
Uploading assets...
Uploading: .DS_Store
Uploading: ._
Uploading: .DS_Store
Uploading: ._
Uploading: .DS_Store
Uploading: ._
```

This doesn't seem right to me. I noticed on https://github.com/ember-cli/ember-cli-deploy/issues/11#issuecomment-62206031 that a suggestion was made to ignore hidden files by default, but no clear follow up was made (as far as I can tell).

How does one go about ensuring these sorts of files aren't uploaded?

lukemelia:
Closes #138

### DIFF
--- a/lib/tasks/assets.js
+++ b/lib/tasks/assets.js
@@ -1,17 +1,12 @@
 var Task                = require('ember-cli/lib/models/task');
 var ConfigurationReader = require('../utilities/configuration-reader');
 var Promise             = require('ember-cli/lib/ext/promise');
-var SilentError         = require('ember-cli/lib/errors/silent');
-var chalk               = require('chalk');
 var fs                  = require('fs');
 var path                = require('path');
-
-var ncp   = Promise.denodeify(require('ncp'));
-var white = chalk.white;
-var green = chalk.green;
-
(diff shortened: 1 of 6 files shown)


lib/tasks/assets.js
```diff
@@ -98,5 +89,16 @@ module.exports = Task.extend({
       stopOnErr: true,
       limit: 2
     });
+  },
+
+  deleteExcluded: function(dir, config){
```
achambers: @lukemelia Based on how you've invoked this function above, it should be called `deletedExcludedFiles`

lukemelia: Thanks, @achambers. Updated the reference. It's pretty bad that our tests are in sufficient to catch such a thing. We really need the integration test that @danshultz started in #117.

achambers: Yeh, I was surprised that the tests didn't catch this.

PR #149, opened 2015-06-07
Add didFail-hook

LevelbossMike:
Some plugins need the ability to react to failures in the deployment
pipeline to for example notify some webservice of the failed deployment.

It is now possible to add a `didFail`-hook to a plugin that gets
executed as soon as a deployment-hook's execution promise rejects.

* add didFail-hook to pipeline
* 'deanonymize' functions in pipeline utility class

### DIFF
--- a/lib/utilities/pipeline.js
+++ b/lib/utilities/pipeline.js
@@ -18,7 +18,7 @@ function Pipeline(hookNames, options) {
     pipelineHooks[hookName] = [];
 
     return pipelineHooks;
-  }, {});
+  }, { didFail: [] });
 }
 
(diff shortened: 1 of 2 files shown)


LevelbossMike:
ping @achambers 

I also tried to cleanup the anonymous functions to make it clearer for others what the pipeline is doing. Method names are pretty long but I think it's more readable that way.

achambers:
Is this good to merge then @LevelbossMike ?

LevelbossMike:
Yes. Worked for my use-case with the slack plugin. So I think this is good to merge

achambers:
Thanks @LevelbossMike

PR #156, opened 2015-06-17
Fix _materialize using wrong this context

jeffhertzler:
This bug was causing custom config options that had defaults to not be used within assets tasks (such as ember-deploy-s3).

See https://github.com/ember-cli/ember-cli-deploy/blob/6b96fe989a24566f80df6a06596678b02d5a45aa/lib/tasks/assets.js#L55
and https://github.com/ember-cli/ember-cli-deploy/blob/6b96fe989a24566f80df6a06596678b02d5a45aa/lib/models/config.js#L44-55

### DIFF
--- a/config/deploy.js
+++ b/config/deploy.js
@@ -2,7 +2,7 @@ module.exports = {
   "development": {
     "store": {
       "host": "localhost",
-      "port": 6379
+      "port": 63790
     },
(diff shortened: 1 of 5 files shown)


lukemelia:
Thanks @jeffhertzler

PR #157, opened 2015-06-17
Git add new adapter to list

leojh:
Add my adapter? :)


leojh:
Sorry have to resubmit

PR #159, opened 2015-06-18
Update video link in README.

blimmer:
This will be useful until 0.5, and is more up-to-date than the current talk.


LevelbossMike:
:+1: cool thx!

PR #162, opened 2015-06-22
[WIP] Implement activate command

achambers:


lib/commands/deploy.js
```diff
@@ -5,18 +5,24 @@ module.exports = {
...
+    };
+
     var PipelineTask = require('../tasks/pipeline');
     var pipeline = new PipelineTask({
       project: this.project,
       ui: this.ui,
       deployEnvironment: commandOptions.environment,
       deployConfigPath: commandOptions.deployConfigFile,
+      commandLineArgs: commandLineArgs,
       hooks: [
```
lukemelia: Not sure how this would work exactly, but what do you think about making these hooks include "willActivate", "activate", "didActivate" only if immediate activation is enabled?

achambers: That makes sense.

PR #166, opened 2015-06-23
Only register initialised hooks

achambers:
Previous to this PR we would loop through all keys on the deploy plugin object and register the functions.  This could be error prone.  Instead we now cycle through the list of hooks that the pipeline has been initialised with and then retrieve them from the deploy plugin


lukemelia:
:+1:

PR #169, opened 2015-06-27
Update ember cli to 0.2.7

samselikoff:

PR #171, opened 2015-06-29
Update ember cli 0.2.7 and fix npm warnings

ghedamat:
this is very similar to https://github.com/ember-cli/ember-cli-deploy/pull/169

the only difference is locking `debug` to version `2.1.1` and locking `glob` to `4.2.2` that seems to prevent any `WARN` from npm.

npm WARN were also causing the tests in https://github.com/ember-cli/ember-cli-deploy/pull/117 to fail,

I'm currently trying to revamp #117 and continue that work but I got blocked by this first

please let me know what you think

p.s

we can probably remove `ember-try` for now? 

/cc @samselikoff @lukemelia


lukemelia:
LGTM. Agreed re ember try. Want to remove that and then we can merge?

samselikoff:
lgtm

ghedamat:
@lukemelia will do right now

ghedamat:
ok, should be good now

lukemelia:
Thanks @ghedamat!

PR #172, opened 2015-07-04
Adds support for “plugin packs” to 0.5.0

lukemelia:
A plugin pack is an addon with the keyword “ember-cli-deploy-plugin-pack”
and one or more dependent addons that are ember-cli-deploy-plugins.

Note that the plugin pack’s dependent addons should be listed as
dependencies in the pack’s package.json, not as devDependencies.

I expect that plugin packs may also implement a config/deploy.js
blueprint that is auto-executed upon `ember install` of the pack.

### DIFF
--- a/lib/tasks/pipeline.js
+++ b/lib/tasks/pipeline.js
@@ -94,6 +94,16 @@ module.exports = Task.extend({
   },
 
   _registerPipelineHooks: function(addon, configuredPluginList) {
+    var self = this;
+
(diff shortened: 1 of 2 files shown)


lukemelia:
The world's first ember-cli-deploy-plugin-pack: https://github.com/lukemelia/ember-cli-deploy-lightning-pack

achambers:
Love it!

PR #175, opened 2015-07-08
Manifest prefix for activate task.

juggy:
Manifest prefix needs to be used for the activate task as well and in all adapters instance.


achambers:
Thanks mate

PR #178, opened 2015-07-09
Use SilentError to log errors when parsing config.

rwjblue:
The original error and stack is already printed to the console, there is no need to throw a "real" error (because the stack within ember-cli-deploy itself is not useful).


achambers:
Thanks @rwjblue

PR #181, opened 2015-07-12
fix failing test on windows caused by path.join returning backslashes…

duizendnegen:
… for the path


lukemelia:
Thanks @duizendnegen!

PR #187, opened 2015-07-28
instructions on installing index/assets plugins

duizendnegen:
tossed in (simple) instructions on picking adapters appropriate to their environment. Not sure whether installation is the right chapter for that. Since it's 0.4.x, I didn't bother too much (let's get installation instructions / user flow through discovering all plugins & commands super clear for 0.5)

not sure about the link formatting by the way... didn't work with GH pages before


samselikoff:
thanks! in the future I think you just want to use links like `"../configuration"`.

PR #192, opened 2015-07-31
list pipeline containing fetchRevisions and displayRevisions

duizendnegen:
Default displayRevisions adapter at https://github.com/duizendnegen/ember-cli-deploy-display-revisions
Sample implementation of listRevisions in https://github.com/duizendnegen/ember-cli-deploy-azure-tables

### DIFF
--- a/lib/commands/list.js
+++ b/lib/commands/list.js
@@ -8,17 +8,28 @@ module.exports = {
   ],
 
   availableOptions: [
-    { name: 'deploy-config-file', type: String, default: 'config/deploy.js' }
+    { name: 'deploy-config-file', type: String, default: 'config/deploy.js' },
+    { name: 'amount', type: Number, default: 10 }
   ],
 
   run: function(commandOptions, rawArgs) {
-    var ListTask = require('../tasks/list');
-    var listTask = new ListTask({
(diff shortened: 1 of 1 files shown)


achambers:
@duizendnegen / @lukemelia Maybe I'm confused but is the default implementation of `displayRevisions` going to be in it's own plugin?  Would it make any sense to add it to [lukemelia/ember-cli-deploy-plugin](https://github.com/lukemelia/ember-cli-deploy-plugin) ?

duizendnegen:
Can't add it to `ember-cli-deploy-plugin` since that's really the base class for any plugin, right? Then `displayRevisions` will trigger a print-out for every plugin you have.
The only possibility is to add it to `ember-cli-deploy` itself (imo), but that'd either be an architectural break (adding it to pre-existing plugins) or ugly in another way (by adding it as a dependency and marking `ember-cli-deploy` itself as a `ember-cli-deploy-plugin-pack`)

Could be mistaken, open for feedback.

achambers:
> Can't add it to ember-cli-deploy-plugin since that's really the base class for any plugin, right? Then displayRevisions will trigger a print-out for every plugin you have

Oh yes, of course.  Good point ;)

PR #194, opened 2015-08-02
Implement aliases to support multiple instances of plugins

dschmidt:
Hey,

this is my initial attempt of implementing the alias feature for plugins.

I tested with the redis-plugin, which does not correctly use the passed in name yet.
See https://github.com/zapnito/ember-cli-deploy-redis/pull/18 .

BR,
Domme

### DIFF
--- a/lib/tasks/pipeline.js
+++ b/lib/tasks/pipeline.js
@@ -46,34 +46,47 @@ module.exports = Task.extend({
   setup: function(){
     var self = this;
     return this._readDeployConfig().then(function(deployConfig){
-      self._plugins = Object.create(null);
-      (self.project.addons || []).forEach(function(addon){
-        self._registerPlugin(addon);
-      });
+      self._installedPlugins = self._discoverInstalledPlugins();
 
-      var aliasHash = self._aliasHash(deployConfig, self._plugins);
+      var plugins;
+      var configuredPluginList = deployConfig.plugins;
+      if (configuredPluginList) {
+        plugins = self._configuredPlugins(configuredPluginList, self._installedPlugins);
+      } else {
+        // no whitelist available, use all autodiscovered plugins
+        plugins = self._installedPlugins;
+      }
 
-      Object.keys(aliasHash).forEach(function(alias) {
-        self._registerPipelineHooks(aliasHash[alias], alias);
+      Object.keys(plugins).forEach(function(pluginName) {
+        self._registerPipelineHooks(plugins[pluginName], pluginName);
       });
 
       return deployConfig;
     });
   },
 
-  _registerPlugin: function(addon) {
+  _discoverInstalledPlugins: function(){
+    var self = this;
+    var installedPlugins = Object.create(null);
+    (this.project.addons || []).forEach(function(addon){
+      self._registerInstalledPlugin(addon, installedPlugins);
+    });
+    return installedPlugins;
+  },
+
+  _registerInstalledPlugin: function(addon, registry) {
     var self = this;
 
     if (this._isDeployPluginPack(addon)) {
       addon.addons.forEach(function(nestedAddon){
-        self._registerPlugin(nestedAddon);
+        self._registerInstalledPlugin(nestedAddon, registry);
       });
       return;
     }
 
     if (!this._isValidDeployPlugin(addon)) { return; }
 
-    this._plugins[this._pluginName(addon)] = addon;
+    registry[this._pluginNameFromAddonName(addon)] = addon;
   },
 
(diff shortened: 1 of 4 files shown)


dschmidt:
While figuring out why two tests were suddenly failing I did reimplement the feature a bit differently, you might or might not like it better ... I just thought it might be worth sharing, so you can choose the least evil :-)

dschmidt:
What's spurious: Why didn't the test fail that ensures that plugins are only added if they are whitelisted?

https://github.com/ember-cli/ember-cli-deploy/blob/0.5.0-dev/node-tests/unit/tasks/pipeline-test.js#L161

lukemelia:
Discussed in Slack but repeating here for onlookers:

If you don't define plugins in config, plugins are autodisovered. We expect this to be the common use case and for whitelisting and aliasing to be a power user move.

lib/tasks/pipeline.js
```diff
@@ -93,11 +92,29 @@ module.exports = Task.extend({
...
-    }
 
-    var deployPlugin = addon.createDeployPlugin({name: name});
+    var configuredNameRegex = /(.*?):(.*)/;
+    var configuredPluginHash = Object.create(null);
+    (deployConfig.plugins || []).forEach(function (pluginName) {
+      var split = pluginName.split(':');
+      if(split[0] === name) {
+        if(split[1]) {
+          configuredPluginHash[pluginName.match(configuredNameRegex)[2]] = split[0];
```
lukemelia: instead of using `configuredNameRegex`, can't we just use `split[1]`?

dschmidt: The regex also supports `redis:foo:bar` which `split[1]` does not. If you don't need that feature - I certainly do not either, I just felt it was one thing less to document for users :-)

lib/tasks/pipeline.js
```diff
@@ -93,11 +92,29 @@ module.exports = Task.extend({
...
+        }
+      }
+    });
+
+    Object.keys(configuredPluginHash).forEach(function(configuredName) {
+      if(configuredPluginHash[configuredName] === name) {
+        self._createDeployPluginsForConfiguredName(addon, name, configuredName);
+      }
+    });
+  },
```
lukemelia: As discussed, we need to support the no-whitelist autodiscovery path. I'm fine with those being two separate branches of code. I think that would make it more obvious what's going on and also help us provide more useful error messages when there is a whitelist entry for a plugin that is not available.

lib/tasks/pipeline.js
lukemelia: JavaScript `.split()` accepts a second parameter giving the maximum number of splits to perform. We could pass `2` to get the behavior you described (though it seems like not a great idea for end-users to do this in the first place).

dschmidt: Oh god why :dancers:

dschmidt: Ah, no. `.split(':', 2);` does something else. It limits the elements in the resulting array to 2 but it does not concatenate all others that are left over.
e.g.
```
"redis:foo:bar:doo:bar:doo".split(':', 2)
["redis", "foo"]
```

but my code returns `foo:bar:doo:bar:doo` as the  plugin alias. `split[1]` would in any case only contain `foo`.

As I said, I don't mind. But maybe we should error out if `split(':').length > 2` even if you don't want to support it.

dschmidt:
Added two tests: one for a single alias defined, one for plugin without alias and two aliases defined.

lib/tasks/pipeline.js
```diff
@@ -78,26 +101,40 @@ module.exports = Task.extend({
...
+
+    for(var i=0, len=aliasList.length; i<len; i++) {
+      var alias = aliasList[i];
+      var split = alias.split(':', 2);
+      var pluginName = split[0];
+      var configuredName = split[1] ? alias.match(configuredNameRegex)[2] : split[0];
+
+      aliasHash[configuredName] = plugins[pluginName];
+
+      if(!aliasHash[configuredName]) {
```
achambers: If we're going to reject because there is a whitelist entry for a plugin that we don't know about, would it be a good idea to maybe collect a list of all the unknown plugins? instead of bailing out at the first instance, would it be better to be able to report to the user that there are "x, y and z" plugins that we don't know about?

dschmidt: Easy enough to implement the collecting bit, I'm just still not too sure how to do the actual error handling and the test.

lukemelia:
Great work @dschmidt. Thank you!

PR #195, opened 2015-08-02
Fix broken/wrong pipeline tests

dschmidt:
While implementing plugin aliases, I ran into two failing tests, which I believe are broken.

`deploy.js` does not whitelist any plugins, thus they can't be registered. And the whitelist can't be added to `deploy.js` because it's also used to test what happens if there's no whitelist at all. Thus I changed the tests to use a different fixture.

What's spurious tho:
Why didn't the test fail that ensures that plugins are only added if they are whitelisted?

https://github.com/ember-cli/ember-cli-deploy/blob/0.5.0-dev/node-tests/unit/tasks/pipeline-test.js#L161

### DIFF
--- a/node-tests/fixtures/config/deploy-for-addons-config-test.js
+++ b/node-tests/fixtures/config/deploy-for-addons-config-test.js
@@ -1,5 +1,5 @@
 module.exports = function(environment) {
   return {
-    plugins: ['foo-plugin']
+    plugins: ['foo-plugin', 'test-plugin']
   };
 }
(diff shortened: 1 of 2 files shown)


achambers:
@dschmidt Can you explain a bit more about why you think that test should not be passing?

dschmidt:
Well, I thought I did. 

Because the tests I changed are *not* failing although their plugin names are not whitelisted. They should fail (like they do with my new implementation) and the test that's supposed to ensure that this can't work should fail as well. Or am I misinterpreting anything here?

dschmidt:
@lukemelia clarified on Slack, if there is no whitelist ... plugins should be auto discovered.

Will adjust my implementation.

PR #196, opened 2015-08-09
Fix bug where fingerprint options in deploy.js don't work

achambers:
Previous to [this commit](https://github.com/ember-cli/ember-cli-deploy/commit/42e3c13f863af3108351411bf93d3d45f2c4b258) a user could specify fingerprinting options in `deploy.js` instead of in their `Brocfile.js`.

Due to the renaming of `DEPLOY_ENVIRONMENT` to `DEPLOY_TARGET` in the above commit, this functionality broke.  This PR fixes that.

So, now users can specify fingerprinting options in `deploy.js` like so:

```javascript
  return {
    fingerprint: {
      enabled: environment === 'production',
      prepend: 'https://some-url.com/'
    }
  };
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -36,7 +36,7 @@ Deploy.prototype.blueprintsPath = function() {
 };
 
 Deploy.prototype.included = function(app) {
-  var deployEnv  = this._deployEnvSetByDeployCommand();
+  var deployEnv  = this._deployTargetSetByDeployCommand();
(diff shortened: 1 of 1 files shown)


lukemelia:
While this functionality ideally belongs in the ember-cli-deploy-build plugin, I'm going to merge this PR since what we have is currently broken.

PR #210, opened 2015-09-02
add setup and teardown hooks to all commands

ghedamat:
after discussion with @lukemelia we're proposing to add two extra hooks that should be implemented in every command (like the configure hook)

`setup`
`teardown`

these are useful for things that have to be setup "around" every command like port-forwarding.

@achambers this seems to be inline with what we discussed as well but let me know your feedback!

### DIFF
--- a/lib/commands/activate.js
+++ b/lib/commands/activate.js
@@ -25,9 +25,11 @@ module.exports = {
       commandOptions: commandOptions,
       hooks: [
         'configure',
+        'setup',
         'willActivate',
         'activate',
-        'didActivate'
+        'didActivate',
+        'teardown'
       ]
     });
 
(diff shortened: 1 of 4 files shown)


ghedamat:
as a sidenote shall we make clearer in the RFC that not all the hooks listed are run in every command?

achambers:
Cool....Nice one.

> as a sidenote shall we make clearer in the RFC that not all the hooks listed are run in every command?

Yeh, it could be clearer couldn't it?

lukemelia:
We should probably break down the pipelines and show the hooks for each.

achambers:
:+1: @ghedamat Want to do that in this PR, or a separate one?

ghedamat:
@lukemelia @achambers you're talking about the readme right?

if so maybe that's a separate PR? just for clarity?

achambers:
I was talking about the RFC. Happy for it to be a separate PR

ghedamat:
@achambers yeah sorry that's what I meant

ok, I'll try to open another PR in the next couple of days, in the meantime you should just be able to merge this

achambers:
Cheers mate

PR #211, opened 2015-09-05
better describe hooks in rfc

ghedamat:


ghedamat:
argh,,sry wrong target branch

Issue #77, opened 2015-03-03
How to use ember-cli-dotenv with ember-cli-deploy?


kcollignon:
I'm using the ember-cli-dotenv package. Has anyone had any luck with including environment variables in to the deploy.js file?

i've tried everything from `process.env.S3_SECRET_KEY` to `ENV.APP['S3_SECRET_KEY`]`

ryanto:
`process.env` will have all your environment variables. 

From the command line you can run: `node -e "console.log(process.env)"` to see what's available.

kcollignon:
Thanks @ryanto for that node command. That has helped me verify that my environment variables aren't being injected in the process env. 

Do you see anything that I might be missing?

Part of environment.js:
```javascript
module.exports = function(environment) {
  var ENV = {
    modulePrefix: 'keying-instructions',
    environment: environment,
    baseURL: '/',
    locationType: 'hash',
    contentSecurityPolicy: {
      'default-src': "*",
      'script-src': "*", 
      'font-src': "*",
      'connect-src': "*", // Allow data (ajax/websocket)
      'img-src': "'*'",
      'style-src': "* 'unsafe-inline'", // Allow inline styles and loaded CSS from http://fonts.googleapis.com
      'media-src': "*"
    },
    EmberENV: {
      FEATURES: {
        // Here you can enable experimental features on an ember canary build
        // e.g. 'with-controller': true
      }
    },

    APP: {
      // Here you can pass flags/options to your application instance
      // when it is created
      S3_ACCESS_KEY: process.env.S3_ACCESS_KEY,
      S3_SECRET_KEY: process.env.S3_SECRET_KEY
    }
  };
```
Part of Brocfile:
```javascript
var app = new EmberApp({
  fingerprint: {
    prepend: 'https://s3.amazonaws.com/mybucket/images'
  },
  dotEnv: {
    clientAllowedKeys: ['S3_ACCESS_KEY', 'S3_SECRET_KEY']
  }
});

lukemelia:
Sounds like we need to look at how ember-cli-dot-env works. Maybe we need to make sure we load after that addon?

kcollignon:
Just for fun, I tried setting the deploy config to something random, but that was in my environment variables to see if it was getting loaded in, and it appeared it did. I received the error `uploadedFilePath is not defined` which I usually get when it just doesn't authenticate, but seemed to prove it was at least having access to process.env

MattNguyen:
[...]

PR #216, opened 2015-09-14
Add support for ember-cli-dotenv

matthewlehner:
Hi!

I'm working through setting up Ember CLI Deploy and noticed #77 talking about using ENV variables through Ember CLI dotenv.

This ensures that `process.env` has the environment variables loaded by Ember CLI dotenv when loading `config/deploy.js` and has no side effects if `ember-cli-dotenv` is not present. :smiley_cat:

I debated whether or not I should open this PR on ember-cli-dotenv, but this allows an easy path for storing and accessing sensitive credentials at deploy time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,9 @@
     "ember-addon",
     "ember-cli-deploy"
   ],
+  "ember-addon": {
+    "after": "ember-cli-dotenv"
+  },
   "dependencies": {
     "broccoli": "^0.13.2",
     "broccoli-gzip": "^0.2.0",


lukemelia:
:+1:

matthewlehner:
Hmph. Had some weird stuff on my end that made this _seem_ like it was working. It doesn't. I'll keep investigating when I've got some more time.

achambers:
One thing to note is that this is baked in to 0.5.0 which is currently in beta and close to release.

PR #220, opened 2015-09-24
Update example-sinatra-app.md

williscool:
same meaning infinitely more readable code and no need for another function


lukemelia:
:+1:

PR #221, opened 2015-09-25
Update how-to-use.md

williscool:
more clear what the command is supposed to be. the old way it seems as if that is a sub command


lukemelia:
Thanks, @williscool

williscool:
awesome thanks for the merge

PR #222, opened 2015-09-26
Add support for postBuildDeployHook

ghedamat:
this adds a config option that will trigger a specific pipeline
every time the project is built.

the use case is to enable pushing assets to a store during development

for example: users of the redis plugin could use this config to get
an updated `index.html` in their redis store:

```
    if (deployTarget === 'development-postbuild') {
      ENV.plugins = ['redis'];
      ENV.build = {
        environment: 'development'
      };

      ENV.redis = {
        keyPrefix: 'edd-cli',
        revisionKey: '__development__',
        allowOverwrite: true,
        type: 'redis',  // this can be omitted because it is the default
        host: 'localhost',
        port: 6379,
        distDir: function(context) {
          return context.commandOptions.buildDir;
        }
      };
    }
```

adding also this to `ember-cli-build.js`
```
 emberCLIDeploy: {                                                               
   configFile: 'config/deploy.js',                                               
   runOnPostBuild: (env === 'development') ? 'development-postbuild' : false,    
   shouldActivate: true                                                          
 },    
```
Todo:
====

* [x] tests?
* [x] check if `postBuild` fires even during deploy..
* [x] add info to readme

Notes:
======

* `index.js` was updated to return an object from the exports instead of a
constructor, this allows the postBuild hook to have the proper context
and hence get access to `this.project` and similar.

* ~~currently the code lives inside `lib/commands/deploy.js`, I'm not
convinced it's a good idea but we do need access to a bunch of stuff
there (hooks, config, the deploy pipeline)~~

* ~~added `runOnPostBuild` option in ENV~~

* ~~maybe a separate pipeline is a good idea? any suggestion on how to
better handle hooks?~~

* ~~will this fire during a "proper" deploy as well? we should prevent it if so (I haven't tested yet)~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,75 @@ After installation, choose [deployment adapters](http://ember-cli.github.io/embe
 
 [Visit the Docs site](http://ember-cli.github.io/ember-cli-deploy/)
 
+## Development workflow
+
+Why do this? `ember-cli-deploy` is an addon for deploying apps. It works fine for cases where your ember-cli app and API app are mostly developed in isolation. However, if you develop your apps more closely or rely on your API app to inject session or initial state data into your ember-cli app's index page, the development process can be cumbersome.
+
+This workflow simply writes your ember-cli app's index to your index plugin on each build so that your API app can read (and possibly modify it) just like it would in production.
+
+During app development use `ember server` just to recompile and serve the assets as usual.
+This way the ember app is served by your backend and there's no need to setup CORS or proxies.
+
+This strategy can be easily accomplished with some configurations parameters.
+
+The following is an example that pushes the `index.html` to redis on each build:
+
+in your `ember-cli-build.js`
+
+```javascript
+  module.exports = function(defaults) {
+    var env = EmberApp.env()|| 'development';
+    ...
+
+    var fingerprintOptions = {
+      enabled: true,
+      ...
+    };
+
+    switch (env) {
+      case 'development':
+        fingerprintOptions.prepend = 'http://localhost:4200/'; // use fingerprinting to prepend your ember server domain path
+      break;
+      ...
+    }
+
+    var app = new EmberApp(defaults, {
+      fingerprint: fingerprintOptions,
+      emberCLIDeploy: {
+        runOnPostBuild: (env === 'development') ? 'development-postbuild' : false, // returns the deployTarget
+        configFile: 'config/deploy.js', // optionally specifiy a different config file
+        shouldActivate: true, // optionally call the activate hook on deploy
+      },
+      ...
+    });
+
+    return app.toTree();
+  };
+```
+
+in `config/deploy.js` you can define a custom pipeline for your deploy env.
+
+```javascript
+  if (deployTarget === 'development-postbuild') {
+    ENV.plugins = ['redis']; // only use the redis pluging
+    ENV.build = {
+      environment: 'development'
+    };
+
+    ENV.redis = {
+      keyPrefix: 'edd-cli',
+      revisionKey: '__development__',
+      allowOverwrite: true,
+      type: 'redis', // this can be omitted because it is the default
+      host: 'localhost',
+      port: 6379,
+      distDir: function(context) {
+        return context.commandOptions.buildDir;
+      }
+    };
+  }
+```
+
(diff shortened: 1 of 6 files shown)


lukemelia:
Idea: the postBuild hook retrieves the config passing a conventionally-named named deploy target (maybe "[environment]-postbuild") and if the returned object has `config.runOnPostBuild === true` then it runs with the retrieved config.

ghedamat:
@lukemelia @elucid updated to a different config format that seems simpler
and extracted logic into separate file

lib/hooks/post-build.js
```diff
@@ -0,0 +1,51 @@
...
+var hooks = ['configure',
+  'setup',
+  'willDeploy',
+  'willBuild', 'build', 'didBuild',
+  'willPrepare', 'prepare', 'didPrepare',
+  'willUpload', 'upload', 'didUpload',
+  'willActivate', 'activate', 'didActivate',
+  'didDeploy',
+  'teardown'
+];
```
ghedamat: we can likely trim this list down for the postBuild hook

lib/hooks/post-build.js
```diff
@@ -0,0 +1,51 @@
...
+  if (!existsSync(fullPath)) {
+    return {};
+  }
+
+  var fn = require(fullPath);
+
+  return fn(deployTarget).runOnPostBuild || null;
+};
+
+module.exports = function(ctx, result) {
```
ghedamat: still need to pass ctx ( the `this` from the hook)

can't use something like `postBuild.call(this, result)` because I need access to the build env to load the proper config variable

ghedamat:
@lukemelia @elucid ping :)
happy to chat about this when you guys have time

ghedamat:
ok in the latest version we support options given from `ember-cli-build.js`

```
 emberCLIDeploy: {                                                               
   configFile: 'config/deploy.js',                                               
   runOnPostBuild: (env === 'development') ? 'development-postbuild' : false,    
   shouldActivate: true                                                          
 },                                                                              
```

ghedamat:
@lukemelia added some tests and took a stab at some docs, happy to get some help (especially in the docs :p)

/cc @elucid

elucid:
Re: last note, this won't run on "proper" deploys unless you explicitly configure it to do so via `emberCLIDeploy.runOnPostBuild`. For instance the example provided in `README.md` would not.

elucid:
I think the `README.md` entry is good but if someone's not looking for this feature in particular it might seem confusing. Might be helpful to snarf some of the content from https://github.com/elucid/ember-cli-redis-proxy/blob/master/README.md

ghedamat:
@lukemelia @elucid done some more. good enough for now?

elucid:
Added some more explicit use cases to `README` and minor wordsmithing

README.md
```diff
@@ -329,6 +329,77 @@ Tagging Adapters need to implement the following method for ember-cli-deploy to
...
 This method has to return a `String`.
 
+## Development workflow
+
+`ember-cli-deploy` is an addon for deploying apps. It works fine for cases where your ember-cli app and API app are mostly developed in isolation. However, there are some cases where using the `--proxy` command line option is inadequate. For example:
+ * authentication happens within your API application
+ * you are progressively updating an app to Ember or developing a hybrid app (i.e. some pages are served by the API application) and you need both applications to work together seamlessly
+ * the API app injects some initial state (e.g. session info or model preloads) into your ember-cli app's index page so that it is available without having to make xhr requests on boot
+
+The development workflow simply writes your ember-cli app's index to your index plugin on each build so that your API app can read (and possibly modify) it just as it would in production.
```
lukemelia: This is confusing because there is no such thing as an "index plugin" in 0.5.0.

lukemelia:
One small doc comment. Address that and I think we're ready to merge. Great work!

elucid:
@lukemelia fixed the reference to `index plugin` and rebased to fix merge conflict.

lukemelia:
Thanks @ghedamat and @elucid. I'm very happy that this will be part of 0.5.0.

ghedamat:
thank you for all the help @lukemelia :)

PR #224, opened 2015-10-05
Update upgrade guide to highlight revision-data instead of revision-key

duizendnegen:


achambers:
Champion. Thanks mate.

PR #225, opened 2015-10-07
Give warning when no plugins are installed

achambers:
Warning looks like this:

image

### DIFF
--- a/lib/tasks/pipeline.js
+++ b/lib/tasks/pipeline.js
@@ -57,7 +57,16 @@ module.exports = Task.extend({
         plugins = self._installedPlugins;
       }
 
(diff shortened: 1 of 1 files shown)


lukemelia:
Probably should ultimately link to: http://ember-cli.github.io/ember-cli-deploy/plugins/

achambers:
True

duizendnegen:
imo we'd at least have to provide a reason why plugins should be installed, like, "no plugins installed - ember-cli-deploy uses plugins to do the actual deployment" (not the right wording, but something in that direction)

achambers:
Ok thanks. Will make some changes.

PR #242, opened 2015-10-17
configuration docs

ghedamat:


docs/v0.5.x/aliasing-plugins.md
```diff
@@ -1,3 +1,18 @@
...
+          "s3-assets:bar-assets",
+          "s3-index",
+          "notify-slack"]
+```
+
+The name specified after the colon will be passed as the `name` property
+of the `options` argument to the addon's `createDeployPlugin` method. Plugins
+should use their name to retrieve configuration values. In this example,
+the foo-assets instance of the s3-assets plugin could have different configuration
+than the bar-assets instance does.
```
lukemelia: This page should include a configuration file snippet showing where `plugins` goes and how the configuration would look.

docs/v0.5.x/configuration-overview.md
```diff
@@ -1,3 +1,112 @@
...
+
+```javascript
+plugins: ["s3-assets", "s3-index", "notify-slack"]
+```
+
+Any plugins not included in the list will not have their hooks executed.
+
+It's also possible to [alias plugins](../aliasing-plugins).
+
+## Plugin developers:
```
lukemelia: Let's keep this as part of an Authoring Plugins section

docs/v0.5.x/aliasing-plugins.md
ghedamat: done just below, let me know if that's what you were thinking

lukemelia: Good, but let's show two redis "instances". Doesn't really make sense without that.

ghedamat: lol right :)

ghedamat: ok @lukemelia done, still a contrived example but should give the idea

lukemelia:
:+1:

PR #247, opened 2015-10-18
add a little info about the naming conventions for config parameters

ghedamat:
/cc @achambers 

unless I missed it we didn't clarify that the config for a plugin is scoped under the plugin name

feel free to kill this if it's unneeded

### DIFF
--- a/docs/v0.5.x/writing-a-plugin.md
+++ b/docs/v0.5.x/writing-a-plugin.md
@@ -76,6 +76,25 @@ module.exports = {
 
 That's seriously about as difficult as it gets. However, read on for some more advanced info to get the most out of your ember-cli-deploy plugin.
 
+### Configuration naming
+By convention, plugin configuration should be kept in `config/deploy.js` and scoped by
+the plugin's name.
+
+E.g. for an `ember-cli-deploy-example` plugin with `name: 'example'`, the configuration might look like:
+
+```javascript
+// config/deploy.js
+module.exports = return function(environment) {
+  return {
+    "example": {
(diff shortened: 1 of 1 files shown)


docs/v0.5.x/writing-a-plugin.md
```diff
@@ -76,6 +76,25 @@ module.exports = {
...
+module.exports = return function(environment) {
+  return {
+    "example": {
+      bucket: "my-app-" + environment,
+      awsAccessKeyId: process.env.AWS_ACCESS_KEY_ID,
+      awsSecretAccessKey: process.env.AWS_SECRET_ACCESS_KEY
+    }
+  };
+};
+```
```
lukemelia: This seems like a weird place for this. What are we trying to communicate to the prospective plugin author?

ghedamat: the idea is that if they don't use the `BasePlugin` and instead they implement their own `configure` hook they should make sure they scope their config under a `key` named as their plugin.

at the same time  even in case of using the `BasePlugin` I didn't feel it was clear how the config  was being looked up in the main config object

that said, if you guys think it is I'm all in for killing this :)

achambers:
@ghedamat Would this be more appropriate in the configuration section of the docs?

achambers:
In fact, let's just put it there altogether. I'll close this now and we can add it there

ghedamat:
@achambers it used to be in the configuration section and @lukemelia suggested to move it here :p 

I'm easy either way, was just trying to clarify how the naming conventions are set up

PR #251, opened 2015-10-19
document didFail hook

ghedamat:


docs/v0.5.x/pipeline-hooks.md
```diff
@@ -99,6 +99,8 @@ didDeploy: --> runs at the end of a full deployment operation.
 teardown: ---> always the last hook being run
 ```
 
+### Additional hooks
+
 In addition, there are a few more specialized hooks that plugins may implement:
```
ghedamat: this hook got here because it was in the rfc but I'm not sure if it ever got implemented

cannot find mention of it anywhere in the code

/cc @lukemelia

lukemelia: The didFail stuff looks good. I will merge. The discoverVersions thing ended up being named fetchRevisions. The description of hooks should probably be underneath the pipelines by command it is documenting. In other words, it's not an accurate description of the hooks for an `ember deploy:list` pipeline.

PR #252, opened 2015-10-19
specialized docs for hooks in different commands

ghedamat:
@lukemelia should be in line with what we discussed.

I've also removed some Notes because they were basically not accurate

let me know what you think!

### DIFF
--- a/docs/v0.5.x/pipeline-hooks.md
+++ b/docs/v0.5.x/pipeline-hooks.md
@@ -8,11 +8,11 @@ Note:
 If multiple plugins implement the same hook they will all be called, if you need to specify the order
 you can use the `plugins: [...]` option in the [configuration](../configuration-overview).
 
-### Hooks by command
+## Hooks by command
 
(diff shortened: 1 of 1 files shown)


docs/v0.5.x/pipeline-hooks.md
```diff
@@ -88,35 +64,76 @@ activate ---------> activate  make a new version live
...
 
 ```
-discoverVersions: --> should return a promise resolving to an array
-                      of version objects. Each version object _must_ have
-                      an `id` property. Each version _may_ have one
+configure: ---> Runs before anything happens
+
+setup: -------> The first hook for every command
+
+fetchRevisions: ----> returns a promise resolving to an hash
```
lukemelia: this doesn't necessarily need to be a promise does it?

ghedamat: almost always it is but yeah it's not necessary, will fix

ghedamat:
@lukemelia updated and tweaked another couple of things

PR #257, opened 2015-10-21
usage overview docu

ghedamat:
I'm not entirely sure what you guys envisioned to be in this page

I took a stab at describing the commands for now

happy to get feedback as usual

/cc @achambers @lukemelia

### DIFF
--- a/docs/v0.5.x/usage-overview.md
+++ b/docs/v0.5.x/usage-overview.md
@@ -1,0 +1,43 @@
+---
+title: Usage Overview
+---
+
+Once you've installed your plugins and configured your deployment strategy you're ready to **deploy**!
+
+ember-cli-deploy currently supports 3 commands, each of them takes a first mandatory argument `<deployTarget>` (i.e. `'production'`) and some optional parameters.
+
+### `ember deploy`
+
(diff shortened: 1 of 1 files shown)


docs/v0.5.x/usage-overview.md
```diff
@@ -0,0 +1,43 @@
...
+
+**Example:**
+
+```
+ember deploy production
+```
+
+### `ember deploy:activate`
+
+Runs the `activate*` hooks that are normally configured by your index plugin and are used to select a `revision` as the primary one to be served by your backend.
```
lukemelia: There is no concept of an "index plugin" any more. Perhaps  "Runs a pipeline with the `activate*` hooks that some plugins implement to make a specific `revision` as the active one to be served by your backend."

lukemelia:
I think this is pretty good overall. Is there anything else you think is necessary @achambers?

docs/v0.5.x/usage-overview.md
ghedamat: right, I guess I still reason in that mentality :D

ghedamat:
@lukemelia should be golden

PR #266, opened 2015-10-24
Add a `--verbose` option to commands, make default output quiet, and …

lukemelia:
…make flag available to plugins via `verbose` property on `ui` object.


ghedamat:
lgtm!

PR #269, opened 2015-10-25
port fingerprint docs from 0.4.x

ghedamat:
adds some minor tweaks as well

/cc @lukemelia (sorry for all the pings)


lukemelia:
Can we verify that the syntax in this example works? I think my ember-cli-build.js file gets the environment differently.

ghedamat:
was about to say... it's on my list for sure, will try to get to it tomorrow

Issue #276, opened 2015-10-28
Add some sort of feedback when --verbose=false


achambers:
Now that we have turned off the logging unless `--verbose` is set some people are feeling like they are missing some form of feedback that something is happening.

Might be worth thinking about adding some sort of simple spinner (maybe like `ember install` uses) when the `verbose` flag is false.

ghedamat:
or maybe just a couple of simple

`starting deploy...`
`deploy done!`

messages?

achambers:
Sure...Anything....It might be nice to use https://www.npmjs.com/package/pleasant-progress like ember-cli does to give a perception of something actively happening.

Dunno...It's just feedback I've been given based on the new "no output by default". While the general inspiration is "if there is no output then everything is good" maybe something small like this might be helpful.

ghedamat:
understood

will see if I can do something on plane :D

jrowlingson:
+1 - I spent more time than I should have last night trying to figure out why my plugin was not logging anything.

ghedamat:
@jrowlingson you're talking about a plugin you're writing right?

for that we definitely need to add mention of the new `verbose` option
on the docs

http://ember-cli.github.io/ember-cli-deploy/docs/v0.5.x/writing-a-plugin/ 

I'll see if I can get to that as well while I fly to embercamp

achambers:
Thanks @ghedamat.

jrowlingson:
@ghedamat Yes, I have started working on the github plugin.

ghedamat:
@jrowlingson glad to hear!

you probably figured it out already but you can do this
https://github.com/ember-cli-deploy/ember-cli-deploy-redis/blob/master/index.js#L69
to make the log always show

otherwise pass the `--verbose` option.

feel free to reach out on `#ember-cli-deploy` on the ember slack if you need help!

lukemelia:
Closed by #280

PR #280, opened 2015-10-29
display progress notification during deploy

ghedamat:
first pass at addressing #276 

Remaining TODOs:

* [x] add commandLine  option to disable progressBar
* [x] add configFile option as well
* [x] ideally figure out a way to resolve rendering issues as seen in the video 
* [x] test this out?
* [ ] update ember-cli-deploy-plugin version in core repos

http://gsnaps.s3.amazonaws.com/screencast_2016-01-08_14-54-36_0_edited.mp4

### DIFF
--- a/lib/commands/deploy.js
+++ b/lib/commands/deploy.js
@@ -9,18 +9,23 @@ module.exports = {
     '<deployTarget>'
   ],
 
+  // note: we can not use `default` from ember-cli because we need to use
+  // settings from .ember-cli config-file as secondary defaults
   availableOptions: [
     { name: 'deploy-config-file', type: String, description: '(Default: config/deploy.js)' },
-    { name: 'verbose', type: Boolean },
-    { name: 'activate', type: Boolean }
+    { name: 'verbose', type: Boolean, description: '(Default: false)' },
+    { name: 'activate', type: Boolean, description: '(Default: false)' },
+    { name: 'show-progress', type: Boolean, aliases: ['p', 'progress'], description: '(Default: true)'}
   ],
 
   run: function(commandOptions, rawArgs) {
     commandOptions.deployTarget = rawArgs.shift();
 
     commandOptions.deployConfigFile = chooseOptionValue(commandOptions.deployConfigFile, this.settings, 'deploy-config-file', 'config/deploy.js');
-    this.ui.verbose = chooseOptionValue(commandOptions.verbose, this.settings, 'verbose');
     commandOptions.activate = chooseOptionValue(commandOptions.activate, this.settings, 'activate');
+
+    this.ui.verbose = chooseOptionValue(commandOptions.verbose, this.settings, 'verbose');
+    this.ui.showProgress = chooseOptionValue(commandOptions.showProgress, this.settings, 'showProgress', process.stdout.isTTY ? true : false);
 
     process.env.DEPLOY_TARGET = commandOptions.deployTarget;
 
@@ -43,6 +48,5 @@ module.exports = {
 
       return deploy.run();
     });
-  },
-
+  }
 };
--- a/lib/models/pipeline.js
+++ b/lib/models/pipeline.js
@@ -22,6 +22,8 @@ function Pipeline(hookNames, options) {
 
     return pipelineHooks;
   }, { didFail: [] });
+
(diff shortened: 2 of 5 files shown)


ghedamat:
shall we make the progress thingy optional if `--verbose` is passed?

lukemelia:
I don't want to rush this into 0.5.0. I'd suggest saving it for 0.5.1.

ghedamat:
works for me :)

seawatts:
Woo hoo this would be awesome!

ghedamat:
current status
http://gsnaps.s3.amazonaws.com/screencast_2016-01-08_14-54-36_0_edited.mp4

we don't count `configure` hooks

we trigger a tick for each of the other hooks, in my demo app we have `10` ticks in total

it's still a bit "choppy" because some hooks are faster than other but I think it's not bad.

feedback is welcome

@lukemelia @achambers @seawatts

ghedamat:
also, this currently runs also for `deploy:list` and `deploy:activate` and ends up looking quite weird.

maybe we should pass an option and do it only for `deploy`?

ghedamat:
once we merge https://github.com/ember-cli-deploy/ember-cli-deploy-plugin/pull/6 and update the relevant plugins we should solve the overlap problem 

http://gsnaps.s3.amazonaws.com/screencast_2016-01-10_22-06-28_0_edited.mp4

ghedamat:
Now with increased niceness
http://gsnaps.s3.amazonaws.com/screencast_2016-01-12_18-19-02_0_edited.mp4

(note that I'll have to add a tweak to `ember-cli-deploy-plugin` for this to work properly)

seawatts:
:+1:

On Tue, Jan 12, 2016 at 3:21 PM, Mattia Gheda <notifications@github.com>
wrote:

> Now with increased niceness
> http://gsnaps.s3.amazonaws.com/screencast_2016-01-12_18-19-02_0_edited.mp4
>
> (note that I'll have to add a tweak to ember-cli-deploy-plugin for this
> to work properly)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ember-cli/ember-cli-deploy/pull/280#issuecomment-171096880>
> .
>



-- 
*Chris Watts* | Software Engineer | cwatts@simplymeasured.com

stefanpenner:
fancy

ghedamat:
@lukemelia @stefanpenner you might know this

what's the consensus on adding an emoji to a command line prompt? from experience I know it likely won't work on linux terminals, so I'm kind of against it but it would be cute (I was thinking of :rocket:)

hhff:
:nail_care:

lukemelia:
@ghedamat no idea of tech details of terminal emoji, but we could probably platform detect, I am :+1: :rocket: :exclamation:

ghedamat:
@lukemelia I think we're ready for a final review, I added the aliases as requested and some test coverage

ghedamat:
ok, to whom is interested, now I consider this ready

only thing to note is that from what I can tell ember-cli help won't print long aliases correctly (maybe this should be a separate PR) 

i.e.

`--show-progress` is aliased to `-p` and `--progress`

but the help is printed as

```
ember deploy <deployTarget> <options...>
  Deploys an ember-cli app
  --deploy-config-file (String) (Default: config/deploy.js)
  --verbose (Boolean) (Default: false)
  --activate (Boolean) (Default: false)
  --show-progress (Boolean) (Default: true)
    aliases: -p, -progress
```

it seems that when using it both `-progress` and `--progress` are being supported
see https://github.com/ember-cli/ember-cli/blob/282641ba662292afb40cb88cdf31557a3e4cc6b7/lib%2Futilities%2Fprint-command.js#L79

lukemelia:
@ghedamat looks great! I have one last idea for your consideration. Shall we only default this to true if the command is being run from a terminal? Seems like we can check `process.stdout.isTTY` ([source](http://stackoverflow.com/questions/7080458/test-whether-the-actual-output-is-a-terminal-or-not-in-node-js))

ghedamat:
@lukemelia done directly in the deploy command so that it's the default

minor problem is: if you use `.ember-cli` option that will override the default, but I don't see why one would force it to "true" as that's the standard behaviour anyway

lukemelia:
Great work @ghedamat!

PR #283, opened 2015-10-29
Tidying up upgrading-plugins guide

samselikoff:


docs/v0.5.x/upgrading-plugins.md
```diff
@@ -2,55 +2,60 @@
...
+
+  4. Use the pipeline hooks to do the actual work:
+      - move uploading logic (from your index/asset adapters) to the `upload` hook
+      - move activation logic (from your index adapter) to the `activate` hook
+
+  5. Instruct your users to [update their config](./configuration-overview) and install [ember-cli-deploy-build](https://github.com/ember-cli-deploy/ember-cli-deploy-build)
+
+  6. Replace `console.log` statements with `this.log` to play nicely with Ember CLI Deploy's messaging
+
+      - Messages passed to `this.log` are displayed only when the verbose flag (`--verbose`) is present. You can pass `{ verbose: true }` as the second param to `this.log` to force your message to always display during pipeline execution.
```
samselikoff: btw, is this true? seems like a confusing api

lukemelia: Basically, yes. Ideas for something clearer?

samselikoff: something like `always: true`? I get `verbose` is saying, "this is a verbose message, always show it", but because of the `verbose` flag it feels like "this is a verbose message, only show it during --verbose runs". Too ambiguous imo.

always: true
verboseOnly: false (default is true)
important: true

achambers: I do agree that this tripped me up. +1 for making it a bit clearer.

PR #290, opened 2015-11-02
Added plugin ordering instructions

duizendnegen:


ghedamat:
:+1:

docs/v0.5.x/writing-a-plugin.md
```diff
@@ -358,3 +358,10 @@ Then once the pipeline hook has run, the deployment context will look like this:
 ```
 
 And every pipeline hook run thereafter will be able to access the `uploadedAt` property.
+
+If you find yourself requiring properties from another plugin in the same hook as it was set, you can specify plugin ordering (and thus property availability) in package.json,
+```javascript
+"ember-addon": {
+   after: "ember-cli-deploy-funky-plugin"
+}
```
achambers: Might be a good idea to put this under a heading as it currently looks like it comes under "adding-data-to-the-deployment-context-object".

Also, is it worth mentioning that by default plugins are loaded in ember-cli's natural order and that the config above is an ember-cli level thing as opposed to ember-cli-deploy, ie it's telling ember-cli which order to load it's addons in.

ghedamat: I thought the same regarding the title, but you can also argue that it's actually part of the discussion on the context, 
maybe subheading instead?

regarding natural order: agreed

achambers: Oh, I see your point. Makes sense.

duizendnegen:
changed to highlight that we're respecting ember-cli ordering

achambers:
Nice work.  Shall we fix this build error in this PR?

duizendnegen:
the build error was already fixed a few prs earlier - could rebase, but it'll disappear on merge

PR #293, opened 2015-11-05
Use npm install when installing plugin base

kpfefferle:
When creating an ember-cli-deploy plugin, the base ember-cli-deploy-plugin module needs to be saved in the `package.json` under `dependencies`, not `devDependencies`. Using `ember install` as previously instructed saves ember-cli-deploy-plugin under `devDependencies` (which we don't want) and switching to `npm install --save` saves ember-cli-deploy-plugin under `dependencies` (which we do want).

This change matches the instructions found in the ember-cli-deploy-plugin README: https://github.com/ember-cli-deploy/ember-cli-deploy-plugin/blob/master/README.md


lukemelia:
Good catch. Thanks @kpfefferie

PR #294, opened 2015-11-05
Require a deploy target to run ember deploy

### LINKED ISSUE #226: Override `production` as the default deploy target.
Our use case: A developer needs to specify a target when deploying to the production.

So `ember deploy production` makes sense when a developer is trying to push to development.
However is `ember deploy` defaults to push to production. 

We need away to override the behaviour. 

Where production is set as the default.
https://github.com/ember-cli/ember-cli-deploy/blob/v0.5.0-beta.2/lib/tasks/pipeline.js#L27

Proposed location for the override.

```js
module.exports = function(deployTarget) {
  var ENV = {
    defaultDeployTarget: 'staging',  //override here
    build: {
      outputPath: 'tmp/dist-assets'
    },
    'revision-data': {
      filePattern: 'do-not-compile.html'
    },
  ......
```

kiwiupover:
Fixes #226 
Pairing with @jheth I enjoyed the time.

### DIFF
--- a/lib/tasks/pipeline.js
+++ b/lib/tasks/pipeline.js
@@ -18,13 +18,16 @@ module.exports = Task.extend({
       throw new SilentError('No ui passed to pipeline task');
     }
 
+    if(!this.deployTarget) {
+      throw new SilentError('You need to provide a deployTarget: `ember deploy production`');
+    }
+
     this._pipeline = this.pipeline || new Pipeline(this.hooks, {
       ui: this.ui
     });
 
     var root = this.project.root;
 
-    this.deployTarget = this.deployTarget || 'production';
     this.commandOptions = this.commandOptions || {};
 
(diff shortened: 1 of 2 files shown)


lukemelia:
LGTM. Thanks @jheth and @kiwiupover!

Ideas about how we should handle docs on this @ghedamat @samselikoff? Even if we merge this now, it won't be in a release right away. It's unlikely to bite anyone badly, but it is technically a breaking change.

ghedamat:
LGTM as well, thanks @jheth @kiwiupover!

@lukemelia, I think should have a discussion about it and about how we want to release,

doesn't sound good to go with 0.6.0 this fast but it's what we *should* do if we wanna put this out.

maybe this simplest thing is to merge it later? we can easily open a PR for the docs as well and we can just use labels to track them..

or we start more complex release flows?

PR #295, opened 2015-11-06
Update wording and code block format for plugin execution order

kpfefferle:
- The live site currently is printing out the `javascript` meant to trigger the syntax on the code block, so I added the extra newline necessary to render this properly
- I also adjusted one word for clarity (changed "specify" to "specifying")


lukemelia:
Thanks!

PR #298, opened 2015-11-09
https does not work on npmsearch

broerse:


achambers:
@broerse I don't have a problem with https.

LevelbossMike:
Does this really not work? I just checked this as well and the page loads normally for me.

duizendnegen:
https also works just fine for me
Closing now, feel free to comment / open another PR if necessary

broerse:
Perhaps it is a Chrome thing. https gives me `NET::ERR_CERT_AUTHORITY_INVALID`

achambers:
I was accessing it in Chrome no problem.

broerse:
It works in Edge so lets skip this PR. It is probably a bug.

PR #308, opened 2015-11-20
fix a couple typos in docs index

csantero:


lukemelia:
Thanks @csantero! :smile:

PR #310, opened 2015-11-25
add on redis key prefix when retrieving current

varblob:
current is just the revision id without the prefix so the example doesn't work.


lukemelia:
Good catch!

PR #316, opened 2015-12-01
Remove temporary fix for broccoli-asset-rev

backspace:
I’m excited to finally be getting to update to 0.5!

Is it true that this temporary fix is no longer necessary? I have `broccoli-asset-rev` version 2.2.0 and it _seems_ like it’s working.


lukemelia:
Thanks @backspace. Also working fine for me on 2.2.0

PR #317, opened 2015-12-01
dotenv files were not loaded in time for deploy tasks

tim-evans:
Our deploy-plugin-pack has been preventing us from deploying to production because the the .env file was being loaded too late. Our default `deploy.js` looks like https://github.com/movableink/ember-cli-deploy-ink-pack/blob/master/blueprints/ink-deploy-config/files/config/deploy.js.

The issue arises when the deploy task pulls configuration before dotenv is loaded: https://github.com/ember-cli/ember-cli-deploy/blob/master/lib/tasks/deploy.js#L9. The variables are applied after the fact in the pipleline task: https://github.com/ember-cli/ember-cli-deploy/blob/master/lib/tasks/pipeline.js#L33

This solution seems to be a good interim solution, but I think a better one would be a singular approach to loading the deploy.js file contents.

### DIFF
--- a/lib/tasks/deploy.js
+++ b/lib/tasks/deploy.js
@@ -1,5 +1,6 @@
 var Task        = require('ember-cli/lib/models/task');
 var existsSync  = require('fs').existsSync;
+var dotenv      = require('dotenv');
 var path        = require('path');
 var PipelineTask = require('../tasks/pipeline');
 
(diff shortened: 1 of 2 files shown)


lukemelia:
Thanks for identifying this issue @tim-evans. I'm going to take a crack at the more singular approach.

lukemelia:
Closing in favor of #319

PR #320, opened 2015-12-03
Add build-plus & CouchDB plug-ins

broerse:


lukemelia:
I'm +1 on adding ember-cli-deploy-couchdb. I think the README for ember-cli-deploy-build-plus needs to do a better job explaining under what circumstances someone would use this in order for us to avoid confusing people by listing it here.

broerse:
Edited this on github. Will try to change it but github seems to want to create a new PR. Will Close and retry.

duizendnegen:
I'll take it one step further, I'd rather see `ember-cli-deploy-build-plus` left out of this mix. I'm also +1 on `ember-cli-deploy-couchdb`, though.

My point here being:
- `ember-cli-deploy-build-plus` sounds like it's better, and it isn't
- it enables `ember-cli-deploy-couchdb` to work in a way that we advice against (relying on a modified build step instead of instructing users how to configure `deploy.js`)

broerse:
`ember-cli-deploy-build-plus` is 100% compatible with `ember-cli-deploy-build` but uses distDir as input and output. That is the only plus. I will add that.

broerse:
Added that: https://github.com/ember-cli/ember-cli-deploy/pull/321

PR #321, opened 2015-12-03
Add build-plus & CouchDB plug-ins

broerse:
(rev 2)

### DIFF
--- a/docs/v0.5.x/plugins.md
+++ b/docs/v0.5.x/plugins.md
@@ -20,7 +20,9 @@ The following plugins have been developed by community members:
 
   - [ember-cli-deploy-azure-blob](https://github.com/duizendnegen/ember-cli-deploy-azure-blob) - deploy assets to Azure Blob
   - [ember-cli-deploy-azure-tables](https://github.com/duizendnegen/ember-cli-deploy-azure-tables) - deploy index.html to Azure Tables
+  - [ember-cli-deploy-build-plus](https://github.com/martinic/ember-cli-deploy-build-plus) - build your app using distDir as input and output
   - [ember-cli-deploy-cloudfront](https://github.com/kpfefferle/ember-cli-deploy-cloudfront) - create a CloudFront invalidation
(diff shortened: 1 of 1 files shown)


duizendnegen:
I will not merge `ember-cli-deploy-build-plus` mentioned in plugins - I want to discourage users from going in this direction for reasons mentioned in #320:
- `ember-cli-deploy-build-plus` sounds like it's better, and it isn't
- it enables `ember-cli-deploy-couchdb` to work in a way that we advice against (relying on a modified build step instead of instructing users how to configure `deploy.js`)

broerse:
@lukemelia @duizendnegen `ember-cli-deploy-couchdb` needs `ember-cli-deploy-build-plus` so there is no good reason not to list it. Other plug-ins can take advantage of the `ember-cli-deploy-build-plus` as well until this issue is fixed. That is why we list it here. Please open this PR again.

broerse:
@duizendnegen I am new to writing modules but when your gave two advice options:

> Releasing ember-cli-deploy-build-plus or providing docs for end users to configure deploy.js appropriately (the latter being my recommendation) is the approach I would advice.

I posted that I would go for the first option. I put in a lot of time to do this as a beginner. I liked to do it but now that I am ready creating `ember-cli-deploy-build-plus` you say the name is wrong and it should not be listed. I asked to open this PR again but please leave it closed. 


 Now I ready and it seems to work fineam done you just close your own advice.

duizendnegen:
Sorry - I really appreciate your efforts and I think the work towards `couchdb` is also good.
I acknowledge I gave the advice that a route to solving would be releasing `ember-cli-deploy-build-plus` - this should have been more nuanced: I believe documentation for end users to configure deploy.js is the only maintainable route for now. Therefore, I want to discourage users relying in `ember-cli-deploy-build-plus`.
For the maintainers of `ember-cli-deploy` this is good learning on how to treat the `context` object, which we will take into account when working on future releases.

I understand this might cause frustration, for which I again apologise. I would underline the maintainable route of writing appropriate documentation for end users, and I would also like to offer my help in doing so.

broerse:
Apologies excepted. Lets solve this together in the future. Tack!

PR #331, opened 2015-12-27
Allow command options to be set in .ember-cli file

kpfefferle:
This is a crude first attempt at using a setting within a project's `.ember-cli` file to get verbose deploy outputs on every deploy by default. It is used by simply adding `"deployVerbose": true` to your project's `.ember-cli` file.

The implementation is inspired by ember-cli's own `usePods` setting (code: https://github.com/ember-cli/ember-cli/blob/3c615aa55dfe6be1adf14ef67e74028b09560138/lib/commands/destroy.js#L68-L70) so that the `--verbose` flag can still be used with `ember deploy` to reverse the default.

Potential areas for feedback:
- [x] Do we like the setting name `deployVerbose`? I believe this could even be nested as `"deploy": { "verbose": true }` if we want to go down the road of putting more of these settings (such as `activateOnDeploy`) in the `.ember-cli` file.
- [x] Is there a style or readability preference between reversing the value of `this.ui.verbose` after it's assigned (as implemented), conditionally assigning the value `this.ui.verbose`, or using a one-line ternary assignment?
- [ ] Testing: how and where should this be tested? I don't see any comparable tests in this project that would be easily adapted to test the input from `.ember-cli`. Feedback or direction here would be especially welcomed.

### DIFF
--- a/lib/commands/deploy.js
+++ b/lib/commands/deploy.js
@@ -1,3 +1,5 @@
+var chooseOptionValue = require('./option-value');
+
 module.exports = {
   name: 'deploy',
   description: 'Deploys an ember-cli app',
@@ -8,14 +10,18 @@ module.exports = {
   ],
 
   availableOptions: [
-    { name: 'deploy-config-file', type: String, default: 'config/deploy.js' },
+    { name: 'deploy-config-file', type: String, description: '(Default: config/deploy.js)' },
     { name: 'verbose', type: Boolean },
     { name: 'activate', type: Boolean }
   ],
 
   run: function(commandOptions, rawArgs) {
     commandOptions.deployTarget = rawArgs.shift();
-    this.ui.verbose = commandOptions.verbose;
+
+    commandOptions.deployConfigFile = chooseOptionValue(commandOptions.deployConfigFile, this.settings, 'deploy-config-file', 'config/deploy.js');
+    this.ui.verbose = chooseOptionValue(commandOptions.verbose, this.settings, 'verbose');
+    commandOptions.activate = chooseOptionValue(commandOptions.activate, this.settings, 'activate');
+
     process.env.DEPLOY_TARGET = commandOptions.deployTarget;
 
     var ReadConfigTask = require('../tasks/read-config');
--- a/lib/commands/activate.js
+++ b/lib/commands/activate.js
@@ -1,3 +1,5 @@
+var chooseOptionValue = require('./option-value');
+
 module.exports = {
   name: 'deploy:activate',
   description: 'Activates a passed deploy-revision',
@@ -6,7 +8,7 @@ module.exports = {
   availableOptions: [
     { name: 'revision', type: String, required: true },
     { name: 'verbose', type: Boolean },
-    { name: 'deploy-config-file', type: String, default: 'config/deploy.js' }
+    { name: 'deploy-config-file', type: String, description: '(Default: config/deploy.js)' }
   ],
 
   anonymousOptions: [
@@ -15,7 +17,10 @@ module.exports = {
 
   run: function(commandOptions, rawArgs) {
     commandOptions.deployTarget = rawArgs.shift();
-    this.ui.verbose = commandOptions.verbose;
+
+    this.ui.verbose = chooseOptionValue(commandOptions.verbose, this.settings, 'verbose');
+    commandOptions.deployConfigFile = chooseOptionValue(commandOptions.deployConfigFile, this.settings, 'deploy-config-file', 'config/deploy.js');
+
     process.env.DEPLOY_TARGET = commandOptions.deployTarget;
 
     var ReadConfigTask = require('../tasks/read-config');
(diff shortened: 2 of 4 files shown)


ghedamat:
@kpfefferle thanks for this!

I'm not a fan of using the flag `--verbose` to reverse the behaviour of the settings in `.ember-cli` but if that's already the convention I'm ok with it.

I'm also :+1: to nest the keys in case we later need to add options and keep things isolated.

for the testing part, I'd assume we'd have to stub the config file, maybe worth looking into how ember-cli does that?

lukemelia:
Thanks @kpfefferle. Cool to see how easy this actually is. I agree with with @ghedamat re: nesting keys, perhaps even under `ember-cli-deploy`?

I'm strongly against `--verbose` ever meaning the opposite of verbose. I'd prefer to override via `--verbose=false` or `--no-verbose`.

Re: testing, the sad fact is that we don't have any unit test coverage at the command level. I would really like us to get there. It doesn't have to this PR though.

kpfefferle:
@ghedamat @lukemelia I made a couple of updates in response to your feedback:
- The `.ember-cli` setting is now formatted as `"ember-cli-deploy": { "verbose": true }`
- This setting sets the default, but can be overridden by an explicitly passed flag i.e. `ember deploy production --verbose=false`

lib/commands/deploy.js
```diff
@@ -15,7 +15,15 @@ module.exports = {
...
-    this.ui.verbose = commandOptions.verbose;
+
+    var hasVerboseSetting = this.settings && this.settings['ember-cli-deploy'] &&
+      this.settings['ember-cli-deploy'].verbose;
+    if (hasVerboseSetting && commandOptions.verbose === undefined) {
+      this.ui.verbose = this.settings['ember-cli-deploy'].verbose;
+    } else {
+      this.ui.verbose = commandOptions.verbose;
+    }
+
```
lukemelia: Maybe clearer as:

```js
this.ui.verbose = commandOptions.verbose;
if (this.ui.verbose === undefined) {
  this.ui.verbose = this.settings && this.settings['ember-cli-deploy'] && this.settings['ember-cli-deploy'].verbose;
}
```

kpfefferle: @lukemelia Your phrasing would change the code's behavior slightly in that it assumes that the `.ember-cli` setting is `true` rather than passing through the setting value (which I suppose would be ok as long as the default remains `false` and the setting is only to override it).

I was attempting to check for (1) the existence of the setting value and (2) the absence of an explicit flag before (3) assigning the actual value of the setting. If you prefer the initial assignment to occur before the conditional, maybe this instead?
```javascript
this.ui.verbose = commandOptions.verbose;
if (this.ui.verbose === undefined && this.settings && this.settings['ember-cli-deploy'] && this.settings['ember-cli-deploy'].verbose) {
  this.ui.verbose = this.settings['ember-cli-deploy'].verbose;
}
```
It's a long conditional, but it retains my original functionality.

lukemelia: @kpfefferle :+1:

kpfefferle:
@lukemelia Updated per your feedback!

lukemelia:
OK, cool. Thanks @kpfefferle. I'm thinking we should also add support for the `--activate` and `--config` options (am I missing any others?) and update the commit message with something the we can paste into the 0.6.x docs when we create them.

kpfefferle:
New commits:
- Copied the reading of the `--verbose` setting from `.ember-cli` to the `deploy:list` and `deploy:activate` commands
- Added similar support for `"ember-cli-deploy": { "activate": true }` to the `deploy` command

Clarification is needed before I add support for the `--deploy-config-file` and `--amount` (for `deploy:list`) flags. These flags [have defaults set in the `availableOptions` settings](https://github.com/ember-cli/ember-cli-deploy/blob/master/lib/commands/list.js#L10-L14) which complicate knowing whether the default value has come from this code or whether the user explicitly set a value that matches the default. I believe that the priority order should be as follows:

1. Explicitly set value (i.e. `--deploy-config-file=config/deploy.js`) always takes top priority.
1. Setting defined in `.ember-cli` takes secondary priority.
1. Default is only used if the above two are not defined.

To accomplish this, I will need to remove the default values from `availableOptions` so that I can check and see if these flags have been explicitly set or not. Then I can accurately apply the `.ember-cli` setting or default value in the correct priority order.

Does this seem like the proper approach to you @lukemelia?

lukemelia:
@kpfefferle yep, that seems reasonable to me. My only question is whether removing the defaults will negatively effect `ember help` output, and whether we can remedy that by including the default in the description

kpfefferle:
@lukemelia I think I've got that all covered now. I added a helper function to choose between the passed value, the `.ember-cli` value, and the default value to clean things up a bit. I also added descriptions to the command options that explain the defaults using the same text that the CLI was generating before.

Do you prefer that all of this get squashed into one commit with a documentation-ready message?

lukemelia:
Yes, looks good! Please do squash @kpfefferle. Thanks for your work on this!

ghedamat:
LGTM (aside from squashing) @kpfefferle thanks!

kpfefferle:
@lukemelia @ghedamat: Squashed! Thanks for your input!

PR #339, opened 2016-01-07
[DOC] Make animated gif loop. Fixes #329

### LINKED ISSUE #329: Non-cyclic GIF
I just noticed that the GIF on the following page does not cycle back to the beginning when it ends:

http://ember-cli.com/ember-cli-deploy/docs/v0.5.x/deployment-context/

Yesterday I was reading that page and I couldn't quite understand the image. Well, I understood it because I am familiar with ember-cli-deploy, but I thought it did a poor job at explaining anything to someone who didn't. Today I was skimming pages and realised it was an animated GIF, and much better an illustration that I originally had thought.

Therefore: I'd suggest that the image is made to run in a loop, for those of us who are slow readers in a small screen and won't notice it moving by the time we scroll past the fold.

lukemelia:


achambers:
:+1:

PR #342, opened 2016-01-09
Allow loading .env files + .env.deploy.<env> files

### LINKED ISSUE #341: Support a base .env.deploy file
Right now, you have to have a separate .env file for each environment. It would be nice to provide default, or shared secrets in a `.env.deploy` file without specifying an environment.

Imagine if your development and staging s3 access keys are the same, right now you have to symlink or copy/paste the file.

I propose the following solution:
1. we load up the `.env.deploy` file
2. we load up the `.env.deploy.<target-environment>` file (keys in here would override the higher level file).

Thoughts?

blimmer:
Fixes #341.


lib/tasks/read-config.js
```diff
@@ -32,14 +32,23 @@ module.exports = Task.extend({
...
+    var dotEnvFilename = '.env';
     var dotEnvFilePath = path.join(root, dotEnvFilename);
 
-    if (existsSync(dotEnvFilePath)) {
-      dotenv.load({
-        path: dotEnvFilename
-      });
-    }
+    // order is important here. vars defined in files loaded first
+    // will override files loaded after.
```
blimmer: I added this comment here because the behavior in dotenv seem backwards to me. I would expect overrides to be loaded last!

ghedamat: interesting, yeah I think dotenv discards already defined vars so you can do overrides using the process ENV

.env
```diff
@@ -0,0 +1,2 @@
+OVERRIDDEN='main-dot-env-flavor'
+SHARED='shared-key'
```
ghedamat: We should see if we can put these into the fixtures directory instead of having them on the top level dir of the project

(they're used only for the tests right?)

blimmer: Yep, I can move them.

lib/tasks/read-config.js
```diff
@@ -20,7 +20,7 @@ module.exports = Task.extend({
 
     this.deployConfigPath  = this.deployConfigPath || 'config/deploy.js';
 
-    if (!existsSync(this.deployConfigPath)) {
+    if (!existsSync(path.join(this.root, this.deployConfigPath))) {
```
blimmer: Since [below](https://github.com/ember-cli/ember-cli-deploy/blob/0706b90ab00771d1382835fe7e67e0a95a48e619/lib/tasks/read-config.js#L47) the root is joined to the config path, this seemed like a sane change.

node-tests/fixtures/.env.deploy.development
```diff
@@ -0,0 +1,2 @@
+ENVTEST=SUCCESS
+OVERRIDDEN='deploy-env-flavor'
```
blimmer: Per @ghedamat 's request, these were moved into the `fixtures/` dir.

ghedamat:
@blimmer LGTM, unless @achambers or @lukemelia have anything to add I think we're good

could you squash the commits please?

Finally once we merge it would be ideal to add this to the guides, if you're up for it a PR against the temporary branch for 0.6.X would be amazing https://github.com/ember-cli/ember-cli-deploy/tree/gh-pages-0.6.x-first-pass

blimmer:
Squash is complete. I will start working on the PR against the docs. Thanks!

lukemelia:
:+1:

ghedamat:
Thanks!

PR #348, opened 2016-01-20
Code cleanup

ef4:
This change:

 - adds jshinting to the test suite
 - makes jshint pass, including some dead code removal
 - fixes the test script definition so it properly expands and finds all tests


ef4:
If you compare the test output of this branch vs the previous successful master build, you can see we were missing many tests before: [before](https://travis-ci.org/ember-cli/ember-cli-deploy/builds/101933732), [after](https://travis-ci.org/ember-cli/ember-cli-deploy/builds/103507318)

lukemelia:
Ahh, the test matching bug just manifested itself in the most recent commit.

Thanks for that and the general cleanup @ef4!

ef4:
That makes sense -- I was surprised that the tests were all passing if they really hadn't been running for a while!

ghedamat:
@ef4 amazing! Thanks for this!

PR #360, opened 2016-02-17
Fix build environment on Configuration Overview

uitgewis:
Update documentation to match actual implementation and blueprints installed initially by ember-cli-deploy.

### DIFF
--- a/docs/v0.6.x/configuration-overview.md
+++ b/docs/v0.6.x/configuration-overview.md
@@ -55,7 +55,7 @@ module.export function(environment){
 
 ### Additional Environments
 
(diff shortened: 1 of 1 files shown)


ghedamat:
Hi @uitgewis, thanks for this
Could you remove the `Gemfile.lock` commit from this PR?

I see it's doing a major update for jekill and I'd rather have that separate (and checked) as there might be some breaking stuff in there

thanks!

uitgewis:
Sure thing! Not a Ruby guru so I probably should leave that alone anyway :stuck_out_tongue:

ghedamat:
thanks!

PR #366, opened 2016-02-24
Move to non-scoped npm package for progress, and newer version.

lukemelia:
- Fixes an issue resulting in crash showing progress bar at certain terminal window sizes


ghedamat:
Nice!

PR #373, opened 2016-03-14
RFC on changing the context object

duizendnegen:
Do not merge until finalized. [Rendered content](https://github.com/ember-cli/ember-cli-deploy/blob/rfc/context/rfc/2016-03-context.md).

Open exploration:
- [ ] Scan the 57 plugins and cluster by functionality, extracting what information is consumed & produced
- [ ] Scan which plugins use functions in the `context` object.

Coding efforts:
- [ ] Immutability of the `context` object
- [ ] `readConfig` semantics, preferably `readContext`

### DIFF
--- a/rfc/2016-03-context.md
+++ b/rfc/2016-03-context.md
@@ -1,8 +1,6 @@
 # Summary
 
-The `context` object, while flexible, is a poor representation of the domain model of deployment. A topic-based `pub/sub` might prove more robust, as might implementing a strict domain model.
-
-More than proposing a definite solution, this should be seen literally as a request for comments.
+The `context` object, while flexible, is a poor representation of the domain model of deployment. A growing ecosystem calls for more guidance. This RFC introduces a stronger domain model, while still maintaining the original flexibility.
 
 # Background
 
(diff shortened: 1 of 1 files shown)


duizendnegen:
Minutes
attendees: @lukemelia [LM] @achambers [AC] @ghedamat [MG] @LevelbossMike [MK] @duizendnegen [PS]

LM: agree with sentiment of changes to context object

MG: full flexibility of context object vs. isolation in single plugins are two extremes, look for middle ground

AC: immutability of context object is desirable, should investigate whether all functions on the context object can be removed

PS: domain model would provide guidance for new plugin authors, catching deprecation

LM: and providing linting

MG: wrap domain model as functions around context, possibly in ember-cli-deploy-base-plugin as test

PS: issues with different plugins having different dependencies on the base plugin, better to ensure validation centralised
 
AC+LM: feature what info plugins consume & produce

LM: this would also enable dry run option

duizendnegen:
Closing this, has been too long & the `context` object hasn't been reported as problematic in a long time, just theoretical problems.

PR #374, opened 2016-03-18
Update dotenv-support.md

skaterdav85:
I noticed that you must have `.env` in addition to a deployment target specific file so figured it might be worth mentioning since it tripped me up for a few minutes.


ghedamat:
@skaterdav85 the `.env` file should not be required actually

maybe you're exposing a bug here or there's a problem with your setup

can you please share the error you're getting?

skaterdav85:
Turns out it was because I called an environment variable `USER` which uses the system's username, not the one I defined in the environment file.

PR #376, opened 2016-03-20
Fix s3-index configuration example in docs/s3-walkthrough

nelsond:
The property name `s3-index` should have quotes, otherwise the JavaScript may not be valid.

This should fix the issue.


ghedamat:
thanks!

PR #377, opened 2016-03-21
add config options for log colors

ghedamat:
as discussed in https://github.com/ember-cli/ember-cli-deploy/issues/358

coordinated with https://github.com/ember-cli-deploy/ember-cli-deploy-plugin/pull/8

### DIFF
--- a/lib/models/pipeline.js
+++ b/lib/models/pipeline.js
@@ -4,8 +4,6 @@ var Promise = require('ember-cli/lib/ext/promise');
 var _       = require('lodash-node');
 
 var chalk = require('chalk');
-var blue  = chalk.blue;
-var red   = chalk.red;
 
 /* This is a generic implementation of a pipeline with ordered, promise-aware hooks,
  * pleasant logging, and failure handling. It should not have any "deployment" domain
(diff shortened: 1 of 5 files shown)


lib/tasks/pipeline.js
```diff
@@ -21,7 +21,8 @@ module.exports = Task.extend({
     }
 
     this._pipeline = this.pipeline || new Pipeline(this.hooks, {
-      ui: this.ui
+      ui: this.ui,
+      config: this.config
```
ghedamat: shall we pass just a subset of `config` instead?

lib/models/pipeline.js
```diff
@@ -14,6 +12,12 @@ var red   = chalk.red;
 function Pipeline(hookNames, options) {
   hookNames = hookNames || [];
   options = options || {};
+  var config = _.merge({
+    log: {
+      info: 'blue',
+      error: 'red'
+    }
+  }, options.config);
```
ghedamat: feedback on the structure for the options is welcome

lukemelia:
- [ ] Need docs

ghedamat:
@lukemelia correct, but that we need to do on `gh-pages` branch

PR #379, opened 2016-03-24
Add basic testing guide

ibroadfo:


duizendnegen:
Seems good.
One thing: "Plugins aren't tested like regular ember addons." Might want to spend a word or two as to why they aren't. But that's optional.

ibroadfo:
agreed, but i don't know why it's the case. :smile:

happy to include an explanation if someone lets me know.

duizendnegen:
It's because the project is a bit of an edge case, it doesn't require a dummy app or an active Ember environment or whatever to do it's functioning - it just relies on stuff being built, then doing whatever with the built files.
In short, it runs in a *node* context, not in an *Ember* context. Maybe that's the best explanation.

Oh and if you could squash the commits into one commit, that would be awesome.

LevelbossMike:
Great pr! Thx a lot! Would be cool if we could also add something about how we usally go about testing external services because a deployment plugin will most likely interact with an external s

ibroadfo:
i'm assuming there'll be more to do so will delay squashing till all done

ghedamat:
LGTM ! thanks @ibroadfo 

@LevelbossMike have all your concerns be addressed?

LevelbossMike:
:+1: LGTM! Thanks again @ibroadfo

ibroadfo:
:grinning: thanks for the notes @duizendnegen @LevelbossMike

PR #383, opened 2016-04-07
Fix the example .ember-cli config

dannyfallon:
Wrap the `verbose` option in double-quotes because without
them it's not valid JSON and we get errors.

```
throw "Error when parsing file in " + path + ". Make sure that you have a valid JSON."
    ^
Error when parsing file in /Users/danny/src/testapp/.ember-cli. Make sure that you have a valid JSON.
```


ghedamat:
thanks @dannyfallon !

PR #385, opened 2016-04-13
Fix config file option

dschmidt:


lukemelia:
Looks good to me. @ghedamat, can you give your blessing?

ghedamat:
LGTM thanks @dschmidt !

PR #388, opened 2016-04-14
Bump lodash version to remove warning

orf:
The first commit appeared to pass the tests when it still had a reference to 'lodash-node' in `lib/models/pipeline.js`, which makes me a bit skeptical.


ghedamat:
thanks @orf for this!

ghedamat:
I thnk it's fine to merge this but we should probably do a beta release first to get ppl the chance to try the lodash change

PR #402, opened 2016-05-26
Fix plugin context merging in the pipeline model

dannyfallon:
Since upgrading to lodash `^4.0.0` in v0.6.2 pipeline context merging
has been broken. When your hook result has an object that exists in
the context already (e.g. `distFiles`) then the results of your hook are
no longer concatenated but instead overwrite the existing values.

For example, if you have `context.distFiles` set to `['file1', file2']`
and your hook returns `{distFiles: ['file3']}` then the resulting distFiles
array is now `['file3', 'file2']`, not `['file1', 'file2', 'file3']` as it
was before.

This is because the `_.merge` function used to merge the contexts was
changed in lodash 4 and is now a simple merger that takes an object and
one or more sources to merge into that object; it no longer takes a custom
merge function. To use a merge function you need to change to use
`_.mergeWith`.

I've made this change and added a test to ensure regression does not
happen in future.

### DIFF
--- a/lib/models/pipeline.js
+++ b/lib/models/pipeline.js
@@ -172,7 +172,7 @@ Pipeline.prototype._notifyPipelinePluginHookExecution = function(ui, fnObject, h
 };
 
 Pipeline.prototype._mergePluginHookResultIntoContext = function(context,result) {
-  return _.merge(context, result, function(a, b) {
+  return _.mergeWith(context, result, function(a, b) {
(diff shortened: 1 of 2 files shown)


LevelbossMike:
LGTM

lukemelia:
Thanks @dannyfallon. Released as 0.6.3.

dannyfallon:
Thank you both :+1:

PR #404, opened 2016-06-02
Fix typo use -> us

keoghpe:


achambers:
Thanks @keoghpe. Could you please edit your commit message to start with `[DOC]`?

ghedamat:
@achambers this targets the `gh-pages` branch so we don't need the prefix.

I really need to do some work on `CONTRIBUTING.md` :/

achambers:
@ghedamat  Ah....hrm....Ok....So, when do we need to do it? When we make changes to the `README` on master or something?

PR #405, opened 2016-06-03
Validate configured plugins before continue with pipeline

zzarcon:
Hi guys, I just added a method to validate the configured plugins before continuing with the pipeline, it basically checks if the installed ones contains the configured ones.

I think this could be really useful for the user to get a nicer feedback instead of just failing after that and showing an unrelated error: **Cannot convert object to primitive value**

It took me some time to me to realize that I was trying to use a plugin that i didn't had installed...

Not sure if we should also throw an error and finish the process somehow... I'm not that familiar about how do you handle this kind of errors here :)

Thanks!

### DIFF
--- a/lib/tasks/pipeline.js
+++ b/lib/tasks/pipeline.js
@@ -112,7 +112,7 @@ module.exports = Task.extend({
     });
 
     if (Object.keys(unavailablePlugins).length !== 0) {
-      this.ui.writeError(unavailablePlugins);
+      this.ui.writeError(Object.keys(unavailablePlugins));
       var error = new SilentError('plugins configuration references plugins which are not available.');
       error.unavailablePlugins = unavailablePlugins;
       throw error;


ghedamat:
Hi @zzarcon

I fully agree and actually I run into this bug yesterday,
the problem is actually here

https://github.com/ember-cli-deploy/ember-cli-deploy/blob/master/lib/tasks/pipeline.js#L115

I believe this line should by

`this.ui.writeError(Object.keys(unavailablePlugins));` because `writeError` can not accept objects (maybe it did in the past? /cc @lukemelia)

this change should give you a meaningful error without having to add extra logic

I have that change in a separate feature-branch that I'm working on but if you want to update this PR and see if that solves the problem I'd be happy to merge it :)

thanks

zzarcon:
You are right @ghedamat, when adding the **Object.keys** to **writeError** the process stops and outputs:



I can update the PR and just change that line, if that's fine for you. Thanks!

achambers:
@ghedamat / @zzarcon Is an error an appropriate way to handle this? Kind of feels a bit strange to blow up just because you have something specified in the `deploy.js`.

Would a warning be more appropriate?

zzarcon:
Good point @achambers, I'll say maybe a warning is more appropriated, but since later the pipeline fails because of trying to use an uninstalled plugin I'm not sure...

ghedamat:
yeah, @achambers @zzarcon leave it as a failure for now, turning it into a warning would require a bunch of restructuring.

Maybe we should leave that to a separate PR and limit this to a bugfix

zzarcon:
@ghedamat PR updated 🚀

PR #406, opened 2016-06-04
add support for requiredHooks in pipelines

ghedamat:
## What changed and Why

A lot of users seem to hit cases where a command is not working properly because of a missing pluging in their setup.

The most common is running `ember deploy:list TARGET` without having `ember-cli-deploy-display-revisions` installed or an equivalent plugin

This PR adds support for a `requiredHooks` configuration parameter for the PipelineTask that we can use to throw an error when none of the available plugins implements the required hook.

## Open Questions:

What other hook should we require to be implemented for `list` and the other commands (if any)?

Shall we change the format of the `hooks` property to be something like
```
hooks: [
  { name: 'fetchRevisions', required: true}
  { name: 'willUpload', required: false }
]
``` 
or any other format?

## Related Issues
https://github.com/ember-cli-deploy/ember-cli-deploy/issues/386

/cc @achambers @lukemelia

### DIFF
--- a/lib/commands/activate.js
+++ b/lib/commands/activate.js
@@ -51,6 +51,9 @@ module.exports = {
           'fetchRevisions',
           'didActivate',
           'teardown'
+        ],
+        requiredHooks: [
+          'activate'
         ]
       });
 
--- a/lib/commands/list.js
+++ b/lib/commands/list.js
@@ -49,6 +49,10 @@ module.exports = {
           'fetchRevisions',
           'displayRevisions',
           'teardown'
+        ],
+        requiredHooks: [
+          'fetchRevisions',
+          'displayRevisions'
         ]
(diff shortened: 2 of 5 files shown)


lukemelia:
@ghedamat I like this solution! Using an array seems fine to me. The `activate` command should likely require the `activate` hook be implemented.

ghedamat:
@lukemelia updated, merge at will if you think it's ok :D

lukemelia:
@ghedamat Looks good. Before we merge, can you amend with a more robust commit message? It should speak for itself in the changelog.

ghedamat:
good call! will do

ghedamat:
updated again @lukemelia

lukemelia:
💯

PR #407, opened 2016-06-06
Ember cli deploy plugin control

achambers:
## What Changed & Why
Explain what changed and why.

## Related issues
Link to related issues in this or other repositories (if any)

## PR Checklist
- [ ] Add tests
- [ ] Add documentation
- [ ] Prefix documentation-only commits with [DOC]

## People
Mention people who would be interested in the changeset (if any)


lib/tasks/pipeline.js
```diff
@@ -125,35 +51,9 @@ module.exports = Task.extend({
...
-
-    return configuredPlugins;
-  },
-
-  _pluginNameFromAddonName: function(addon) {
-    var pluginNameRegex = /^(ember\-cli\-deploy\-)(.*)$/;
-    return addon.name.match(pluginNameRegex)[2];
+  _plugins: function() {
+    var registry = new PluginRegistry(this.project, this.config, this.ui);
+    return registry.plugins();
```
ghedamat: @achambers for when you rebase, I think the call to the validation function might go here (?)

lib/models/plugin-registry.js
```diff
@@ -0,0 +1,205 @@
...
+
+  _warnIfConfigReferencesUninstalledAddons: function(list, validAddons, configKey) {
+    var addonNames = Object.keys(validAddons);
+
+    var unknownAddons = list.filter(function(name) {
+      return addonNames.indexOf(name) === -1;
+    });
+
+    if (unknownAddons.length) {
+      var message = chalk.yellow('You have referenced the following uninstalled addons in `' + configKey + '`:\n');
```
ghedamat: nitpick but

is `uninstalled` the right working here? I assume this is replacing the old
`plugins configuration references plugins which are not available.` message right?

maybe should be `You have referenced the following addons in ... and they're not installed` or something similar

lib/models/plugin-registry.js
```diff
@@ -0,0 +1,205 @@
...
+
+    this._warnIfConfigReferencesUninstalledAddons(Object.keys(disabledMap), addons, 'config.pipeline.disabled');
+
+    return disabledMap;
+  },
+
+  _buildRunOrderMap: function(config, addons) {
+    this._warnIfConfigReferencesUninstalledAddons(Object.keys(config), addons, 'config.pipeline.runOrder');
+
+    return config;
```
ghedamat: I'm not sure I fully understand why this is needed sry

ghedamat:
@achambers had a first look, very nice! just a couple of tiny comments so far.
I guess I'd have to see the docs to be sure of the scope of this but I assume it's still what you outlined in https://github.com/ember-cli-deploy/ember-cli-deploy/pull/349#issuecomment-223133205 correct?

lib/models/plugin-registry.js
achambers: @ghedamat, you mean the actual function? Mainly for consistency and to be intention revealing. We need to generate the alias map, disabled map and run order map. The first two need some massaging of the config. The run order map doesn't but this function just kind of keeps the consistency, and is a place to add the warning too I guess. 

That's kind of the reasoning I guess

achambers: If you see lines 16-23 Ive tried to lay out the steps needed to return the plugins as there is quite a bit of logic that happens.

lib/models/plugin-registry.js
achambers: Yeh, I'm not precious about the copy. Happy to change it to whatever you feel describes things best

lib/models/plugin-registry.js
ghedamat: ok. this is what I thought
I saw the steps lines but wanted to confirm..
I guess we can leave it in case we ever need to do something w/it later and it's still a good place for the warning
👍

achambers: That was my thinking ;)

lib/models/deprecated-plugin-registry.js
```diff
@@ -0,0 +1,50 @@
...
+    if (this._isUsingOldPluginControl()) {
+      if (this._isUsingNewPluginControl()) {
+        var message = 'Use of the old and new plugin controls simultaneously does not make sense.\n' +
+          'Please use the new plugin controls\n' +
+          'See https://some-url for more information\n';
+        throw new SilentError(message);
+      } else {
+        ui.writeError('Use of the `config.plugins` property has been deprecated.\n');
+        ui.writeError('Please use the new plugin controls\n');
+        ui.writeError('See https://some-url for more information\n');
```
lukemelia: TODO: update the URL here

lib/models/plugin-registry.js
```diff
@@ -0,0 +1,205 @@
...
+    var self          = this;
+    var graph         = new DAG();
+    var sortedPlugins = [];
+
+    try {
+      plugins.forEach(function(plugin) {
+        var before = self._asArray(runOrderMap[plugin.name], 'before').concat(self._asArray(plugin, 'runBefore'));
+        var after = self._asArray(runOrderMap[plugin.name], 'after').concat(self._asArray(plugin, 'runAfter'));
+        graph.addEdges(plugin.name, plugin, before, after);
+      });
```
lukemelia: We previously considered ember addon DAG order. Do we intend to drop that altogether?

lib/models/plugin-registry.js
```diff
@@ -0,0 +1,205 @@
...
+    var self = this;
+    return addons.reduce(function(validAddons, currentAddon) {
+      var name = self._pluginNameFromAddonName(currentAddon);
+
+      if (self._isValidDeployPlugin(currentAddon)) {
+        validAddons[name] = currentAddon;
+      }
+
+      if (self._isDeployPluginPack(currentAddon)) {
+        var nestedAddons = self._discoverValidAddons(currentAddon.addons);
```
lukemelia: I <3 recursion.

lib/models/deprecated-plugin-registry.js
achambers: Yep, the one failing test is reminding me of exactly that

lib/models/plugin-registry.js
achambers: @lukemelia remind me what you mean by ember addon DAG order?

achambers: Is that the ability to specify it in the package.json? Like we do in the s3 plugin? If so, I vote for ditching that and keeping it all in the deploy.js

lib/models/plugin-registry.js
ghedamat: cool, luke approves so feel free to take what I have up there

lib/models/plugin-registry.js
ghedamat: @lukemelia @achambers I might be misunderstanding but, my concern here would be that we still want defaults to come from the plugins right?

is this getting rid of it?

ghedamat: nevermind, I think you're talking about the "implicit" order from embercli when not the user nor the plugin exposes preferences..

it's remote, but if we ditch that there's the possibility of regressions as we might be changing existing behaviour no?

achambers: @ghedamat , ok so here is my thinking/understanding:

The default order of plugins will be the order that EmberCLI loads them in, just as happens today.  eg, if the plugins are loaded in this order: `['foo', 'bar', 'baz']`, then they will be executed in this order.

An author of a plugin can also specify the plugin to `runBefore` or `runAfter` some other plugin. These will be merged into the initial default order. eg, for the above default order, if the *bar* plugin is specified to `runAfter` *baz* then the execution order will be: `['foo', 'baz', 'bar']`.

A user can specify in config if plugins run `before` or `after`. These will be merged into the above order as well.

### Question
One question that comes to mind. If the *foo* plugin has an author specified `runAfter` of *bar*. And then the user adds this to the their config:

```javascript
pipeline: {
  runOrder: {
    foo: { after: 'baz' }
  },
},
```

Does the config trump the author provided an effective make the run order of *foo* equal to `[''baz]` or should it merge the author provided one and the user provided one and make the run order of *foo* equal to `['bar', 'baz']`?

ghedamat:
@achambers could you close and reopen to target `master` ?

lib/models/plugin-registry.js
```diff
@@ -0,0 +1,278 @@
...
+    return this._addonHasKeyword(addon, 'ember-cli-deploy-plugin-pack');
+  },
+
+  _isValidDeployPlugin: function(addon) {
+    return this._addonHasKeyword(addon, 'ember-cli-deploy-plugin') && this._addonImplementsDeploymentHooks(addon);
+  },
+
+  _pluginNameFromAddonName: function(addon) {
+    var pluginNameRegex = /^(ember\-cli\-deploy\-)(.*)$/;
+    return addon.name.match(pluginNameRegex)[2];
```
ghedamat: on an initial test in a real project this ends up erroring badly

I think you need to guard this or maybe there's something else going on?

achambers: Thanks mate. Will have a look tonight

achambers:
@ghedamat, I'm thinking I'll just force push my branch to the original branch that the original PR was based off as I'd like to keep @ef4 's PR and the discussion that happened around it. 

Cool with that?

ghedamat:
works for me, 

if then you can also rebase that branch it would be golden

achambers:
Yep. Of course

lukemelia:
Code and test coverage look good. I find the the chained functional style in the plugin registry a little hard to follow as a reader, but I think it's fine when paired with the comprehensive test coverage.

lib/models/plugin-registry.js
lukemelia: @achambers I think you just need to move line 47 `var name = self._pluginNameFromAddonName(currentAddon);` to inside the following `if` block. Testing on one of my apps works with that change. I think we need to add a test simulating the presence of an ember-cli addon that is not an ember-cli-deploy plugin,

achambers:
Hmm. Bummer. Because I did it they way to make it easier to understand. There is quite a lot of logic going on to get the list of plugins so that style was intended to be more intention revealing. To tell a story about how we get from a list of adding to a list of plugins. 

Any thoughts on how I make it easier for the reader then? (Not something that's worth changing on this PR now but would like to follow up with it as this is there bulk of where the logic is so would like to make it easier to follow)

achambers:
PS - Thanks for reviewing. I'll fix up that error and at that test.

lukemelia:
@achambers not sure -- as you say, there is a lot of logic. The use of the intention-revealing method names is good. The names reveal "what" is happening. Perhaps code comments within the body of each of those methods could describe "how" it is happening?

achambers:
@lukemelia I guess I was trying to write it such that you don't really need to know how....The *intention* was that from this list: 

```
 var allEmberCliAddons = this._project.addons || [];
 var validAddons       = this._validAddons(allEmberCliAddons);
 var aliasMap          = this._buildAliasMap(validAddons, this._aliasConfig);
 var disabledMap       = this._buildDisabledMap(aliasMap, this._disabledConfig);
 var installedPlugins  = this._installedPlugins(validAddons, aliasMap);
 var runOrderMap       = this._buildRunOrderMap(this._runOrderConfig, aliasMap, validAddons, installedPlugins);
 var enabledPlugins    = this._applyDisabledConfig(installedPlugins, disabledMap);
 var orderedPlugins    = this._applyRunOrderConfig(enabledPlugins, runOrderMap);
```

you can understand the following:

1) We have a list of ember-cli addons that are installed
2) We convert that to a list of valid ember-cli-deploy addons
3) We generate the alias map
4) We generate the disabled bap
5) Based on the alias map we work out what the installed plugins are
6) We generate the run order map
7) We apply the disabled map to determine only the enabled plugins
8) We apply the run order map to determine the ordered list of plugins

I was hoping, as a reader of the code, you wouldn't really need to know how each of those things worked.

*HOWEVER*, I have had a couple of things on my mind in regards to things.

1) I tried to make a distinction between an installed addon (an ember-cli addon - whether it be an ecd one or not) and an installed plugin (the object returned from `createDeployplugin` for which there could be multiple instances for an addon based on aliases). I tried to make the distinction obvious but I'm not sure I quite got there. This still may not be obvious.

2) The return values of the _build***Map functions is not immediately obvious. I constantly found myself getting confused about what the object I was working with looked like. This is partly just the nature of things but I did toy with the idea of commenting what those functions took and the structure of what they returned.

3) The reasoning of why we need the map objects (aliasMap, disabledMap etc etc) isn't immediately obvious. It makes sense when you know but maybe this is the sort of thing that could also be documented in comments.

Are the 3 points described above more what you mean about the "how" it's working? 

I'm certainly happy to make some changes to make it more understandable as this is where the bulk of the work goes and, before pulling the logic out in to this object, I continuously got confused about what was going on in the pipeline if I wasn't frequently in the code....The same could still happen with this so it's in our own interests to make it easy to understand for when we find ourselves back here.

achambers:
I'll add some comments to the functions and push and see how we get on

achambers:
PS - I just ran some tests with some real deploys and all seems good to me once that bug is fixed. Anyone else see any major broken things?

lukemelia:
@achambers you summarized the challenges I had very well. Comments describing return value structure could help a bunch. For naming, I would expect the things you mentioned to be called "addon", "plugin", and "pluginInstance".

achambers:
Okie doke. Will apply the above and see what we come up with.

achambers:
Not quite sure what the deal is with these failing tests....Can't seem to replicate on my local env.... Yu guys seen this? /cc @ghedamat @lukemelia

ef4:
Thanks for pushing this over the finish line!

lukemelia:
👏 👏 👏 👏 👏 👏 👏 👏 👏

PR #418, opened 2016-07-12
[WIP] Add ability to --dry-run

achambers:
## What Changed & Why
This PR adds the `--dry-run` flag to the ember-cli-dpeloy commands. This flag will enable a dry run of the pipeline execution, essentially, switching out the plugin hook implementations with an empty function. This will mainly be used to to test the execution order of plugins.

## Related issues
N/A

## PR Checklist
- [ ] Add tests
- [ ] Add documentation
- [ ] Prefix documentation-only commits with [DOC]

## People
@ghedamat / @lukemelia


ghedamat:
👍 /cc @lukemelia

lukemelia:
What do you think a developer would expect from running with `--dry-run`? Personally I would think of rsync's behavior, where all actions occur except the ones that would mutate state of the machines involved, which just output what would have happened.

The proposed implementation would do a lot less than that, so perhaps we should a) let it do more, b) call it something else.

For (a), maybe there could be a way to allow plugins to opt-in to providing their own dry-run behavior?

PR #443, opened 2016-12-07
Pass on negative exit codes on failure

duizendnegen:
## What Changed & Why
Return non-zero exit code when pipeline fails.

## Related issues
https://github.com/ember-cli-deploy/ember-cli-deploy/issues/417

## People
@v-shan

Demo:
exit-code

### DIFF
--- a/lib/models/pipeline.js
+++ b/lib/models/pipeline.js
@@ -126,7 +126,7 @@ Pipeline.prototype._abortPipelineExecution = function(ui/*, error */) {
     ui.write(this.logInfo('|\n'));
   }
   ui.write(this.logError('Pipeline aborted\n'));
-  return RSVP.reject();
+  return -1;
 };
 
 Pipeline.prototype._notifyPipelineCompletion = function(ui) {
@@ -134,6 +134,7 @@ Pipeline.prototype._notifyPipelineCompletion = function(ui) {
     ui.write(this.logInfo('|\n'));
     ui.write(this.logInfo('Pipeline complete\n'));
   }
+  return 0;
 };
 
 Pipeline.prototype._notifyPipelineHookExecution = function(ui, hookName, context) {
--- a/node-tests/unit/models/pipeline-test.js
+++ b/node-tests/unit/models/pipeline-test.js
@@ -73,7 +73,8 @@ describe ('Pipeline', function() {
         hooksRun.push('2');
       });
       return expect(subject.execute()).to.be.fulfilled
-        .then(function() {
+        .then(function(returnCode) {
+          expect(returnCode).to.eq(0);
           expect(hooksRun.length).to.eq(2);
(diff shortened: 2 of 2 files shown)


lukemelia:
Reading this through makes me wonder if this should be fixed upstream in ember-cli. A rejected promise seems like it should result in a non-zero exit code.

lukemelia:
I queried the #dev-ember-cli channel on slack about this:

```
lukemelia [10:20 AM]  
Should ember-cli return a non-zero exit code for you if your command’s `run` method returns a promise that becomes rejected?

rwjblue [10:20 AM]  
@lukemelia yes, I think it should

tbieniek [10:20 AM]  
@lukemelia I'd say yes

rwjblue [10:20 AM]  
does it not?

lukemelia [10:21 AM]  
Apparently, not? Seems like @duizendnegen needed to do this to make ember-cli-deploy return a non-zero exit code: https://github.com/ember-cli-deploy/ember-cli-deploy/pull/443

tbieniek [10:21 AM]  
@lukemelia what CLI version?

rwjblue [10:22 AM]  
@lukemelia I can totally believe this was broken in the past, but we should confirm on current

tbieniek [10:22 AM]  
there are some exit code issues at the moment, so maybe verify with the last stable release

rwjblue [10:22 AM]  
@tbieniek I don't think this is directly related to current thing FWIW

[10:22]  
this "feels" like a work around for older issues

tbieniek [10:22 AM]  
¯\_(ツ)_/¯

[10:22]  
just a thought

rwjblue [10:22 AM]  
I'm testing locally too (with an empty command)

lukemelia [10:22 AM]  
OK will do. I see a test case for ember build returning a non-zero exit code on failure, but it does that by resolving with an integer

rwjblue [10:24 AM]  
@lukemelia right, but notice how `logError` works here:

https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L130
https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L160
https://github.com/ember-cli/ember-cli/blob/master/lib/cli/cli.js#L220

[10:25]  
tldr; it `return 1` in the `.catch` block

lukemelia [10:25 AM]  
hmm, so that should be sufficient for any failed promise

rwjblue [10:26 AM]  
and that propagates through to: https://github.com/ember-cli/ember-cli/blob/master/bin/ember#L32-L35 (edited)

lukemelia [10:27 AM]  
OK, that’s helpful. We’ll do some more testing to figure out if this PR is needed or not.
Thanks!

rwjblue [10:27 AM]  
ya, good luck please let me know
s/me/us/
```

duizendnegen:
Works also without since `ember-cli` >= `0.2.11`: https://github.com/ember-cli/ember-cli/commit/b8ebbcf1395b60711597a1eaedcd5f2d27e790d7

PR #447, opened 2016-12-19
Update outdated link in afterInstall message

marpo60:
* Use ember-cli-deploy.com instead of ember-cli-deploy.github.io

* Remove version from the link
The downside of this is that getting to the quick section is more than one copy paste away
The good part is that this blueprint will not demand changes after changing versions

### DIFF
--- a/blueprints/ember-cli-deploy/index.js
+++ b/blueprints/ember-cli-deploy/index.js
@@ -9,6 +9,6 @@ module.exports = {
     // to us
   },
   afterInstall: function(options) {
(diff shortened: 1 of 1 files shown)


ghedamat:
@marpo60 thanks for this!

Given that we're about to release 1.0 I'm wondering if we should punt on this pr and tweak it to point to the 1.0 quickstart docs 


@lukemelia thoughts?

lukemelia:
@ghedamat yes, let's point to 1.0.0 docs

marpo60:
There is no http://ember-cli-deploy.com/docs/v1.0.x/quickstart/ yet, so I pointed to http://ember-cli-deploy.com/docs/v1.0.0-beta/quickstart/

lukemelia:
Thanks @marpo60!

PR #455, opened 2017-02-14
Tweak examples in Additional Environments section.

rwjblue:
Using `environment` as a variable name here makes the example slightly
confusing.  This updates to use `deployTarget` throughout, and makes it
clear which file the snippet in question references.


achambers:
Legendary. Thankyou sir @rwjblue

PR #458, opened 2017-02-20
Fix test commands to include all tests

lolmaus:
## What Changed & Why
I've noticed that `npm run test` only runs tests from `index-test.js`.

Here's the latest build on `master` as of Feb 20:
https://travis-ci.org/ember-cli-deploy/ember-cli-deploy/builds/201523239

For some reason, the glob `**` works as a single asterisk `*`, meaning exactly one level of depth. I've updated Mocha, but it didn't resolve the problem. `--recursive` didn't help either.

So I updated the test command to include paths with two distinct levels of depth.

## Bonus
Bumped `mocha` version.

## People
@lukemelia @LevelbossMike

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
     "test": "tests"
   },
   "scripts": {
-    "test": "node node_modules/mocha/bin/mocha node-tests/**/*-test.js",
-    "autotest": "node node_modules/mocha/bin/mocha --watch --reporter spec node-tests/**/*-test.js"
+    "test": "node node_modules/mocha/bin/mocha node-tests/unit/*-test.js node-tests/unit/*/*-test.js",
+    "autotest": "node node_modules/mocha/bin/mocha --watch --reporter spec node-tests/unit/*-test.js node-tests/unit/*/*-test.js"
   },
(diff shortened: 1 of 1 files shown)


ghedamat:
@lolmaus that is definitely interesting, I'm quite sure it used to work in the past...

see https://travis-ci.org/ember-cli-deploy/ember-cli-deploy/builds/133134377 for example

I wonder what changed

I'd rather avoid listing all directories because we could miss one when we add more test.. I'll do a couple of quick experiments and if they fail will merge

in the meantime thanks!

ghedamat:
@lolmaus my suspicion was confirmed, there's a better fix
https://github.com/ember-cli-deploy/ember-cli-deploy/pull/460

somehow travis stopped adding the quotes that made the glob work so now we have to force them

might be good to look at the other ember cli repos to see if the issue is happening there as well

lolmaus:
@ghedamat It happened to me locally too! I'm on Linux.

ghedamat:
werd,

try to see if master is now working correctly with my fix if you can

thanks!

lolmaus:
@ghedamat 

image

ghedamat:
Nice!

PR #467, opened 2017-04-05
Fix upgrade guide to have correct syntax for aliasing plugins

calvinlough:
## What Changed & Why
The guide for upgrading from 0.6 to 1.0 contained incorrect syntax for aliasing plugins. The correct syntax is found here: http://ember-cli-deploy.com/docs/v1.0.x/including-a-plugin-twice/


lukemelia:
Thanks @calvinlough!

PR #469, opened 2017-04-29
Allow more flexible plugin naming

duizendnegen:
## What Changed & Why
Using `deployjs`, addons shouldn't need to start with `ember-cli-deploy` necessarily.

## PR Checklist
- [x] Add tests
- [x] Fix tests

### DIFF
--- a/lib/models/plugin-registry.js
+++ b/lib/models/plugin-registry.js
@@ -273,8 +273,11 @@ module.exports = CoreObject.extend({
   },
 
   _pluginName: function(addon) {
-    var pluginNameRegex = /^(ember\-cli\-deploy\-)(.*)$/;
-    return addon.name.match(pluginNameRegex)[2];
+    if(addon.name.indexOf('ember-cli-deploy') > -1) {
+      var pluginNameRegex = /^(ember\-cli\-deploy\-)(.*)$/;
+      return addon.name.match(pluginNameRegex)[2];
+    }
+    return addon.name;
   },
 
   _addonHasKeyword: function(addon, keyword) {
--- a/node-tests/unit/models/plugin-registry-test.js
+++ b/node-tests/unit/models/plugin-registry-test.js
@@ -90,6 +90,26 @@ describe('Plugin Registry', function() {
     expect(logOutput[3]).to.eq('Visit http://ember-cli-deploy.com/plugins/ for a list of supported plugins.');
   });
 
+  it('accepts plugins with names not starting with ember-cli-deploy and renames those starting with', function() {
+    var validPlugin = makePlugin('foo');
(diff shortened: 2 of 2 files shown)


lukemelia:
@duizendnegen The original motivation for the naming scheme was to essentially use npm to avoid plugin name collisions, since the system has no recourse for dealing with this. Do you have thoughts on this?

duizendnegen:
Mmm, I don't think this change affects this either way. Currently this simply throws if a plugin doesn't start with `ember-cli-deploy-` in the name field (see the first commit, the failing test).
I thought it was strange to have a plugin that builds Angular called `ember-cli-deploy-deployjs-angular-build`. It's just naming, but it feels awkward and end-user perceived value (developers being the end users here) is key.

I'd be happy to scrape the plugin name from `package.json` instead of from the `name` field on the plug-in, to enforce stronger plugin name uniqueness. Still, the same issue here will emerge, so I'd rather get this in and then move on using `package.json` as a source.
To add a warning, this is backward incompatible in the scenario that the `name` field on the `ember-cli-deploy-plugin` is different than that in `package.json`. Additionally, it will still create conflicts if a plugin is called `ember-cli-deploy-deployjs-plugin-build` and another one `deployjs-plugin-build` (but I think this is a corner case we can safely disregard / provide a fair warning for / give feedback to as it happens)

duizendnegen:
After discussion with @lukemelia on Slack:
this change *only* makes it possible for plugin names to start with something else than `ember-cli-deploy-`. This is the basis to move with further de-duping of the `name` field.

A roadmap to implement this further and reduce naming duplication could look like the following:
- we could pull the deploy plugin name from `package.json` as a function (named `pluginName`);
- adding `defaultPrefix` can chip off `ember-cli-deploy-` portions of the plugin names is configured for the project appropriately.
This might result in a situation that newer versions of `ember-cli-deploy-plugin`s are not compatible with older versions of `ember-cli-deploy`, so plugins that adopt the `pluginName` directive should have a major version increment.

/cc @ghedamat @achambers :+1: / :-1:

ghedamat:
I'm 👍 with this change, and like the `pluginName` idea as part of the base plugin that supersedes the `name` property.

achambers:
@duizendnegen I'm 👍 for this PR.

One thing to keep in mind for the future points you mentioned, and I'm not clear on your thinking here so might be of kilter a little, but in terms of the `name` property, if you are referring to the prop we pass in to `createDeployPlugin({ name: someName })`, just be mindful that that `name` property is not necessarily the plugin name, but will be an alias as defined in config and passed in by the plugin registry.

Just something to keep in mind. It may or may not affect what you're thinking.

PR #473, opened 2017-06-20
Reject undefined config

duizendnegen:
this happens after not returning anything from the function in `config/deploy.js`

### DIFF
--- a/lib/tasks/read-config.js
+++ b/lib/tasks/read-config.js
@@ -59,8 +59,8 @@ module.exports = Task.extend({
   _readDeployConfig: function() {
     var deployConfigFn  = require(this.absoluteDeployConfigPath);
     var deployConfig = deployConfigFn(this.deployTarget);
-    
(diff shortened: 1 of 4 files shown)


duizendnegen:
added tests & fixed Windows `npm test` (it only wants double quotes, don't ask me why)

ghedamat:
I like!

lib/tasks/read-config.js
```diff
@@ -58,6 +58,11 @@ module.exports = Task.extend({
 
   _readDeployConfig: function() {
     var deployConfigFn  = require(this.absoluteDeployConfigPath);
-    return RSVP.resolve(deployConfigFn(this.deployTarget));
+    var deployConfig = deployConfigFn(this.deployTarget);
+    
+    if(deployConfig === undefined) {
```
ghedamat: maybe just `if (deployConfig) {` ?

also space 👮

PR #487, opened 2017-08-10
Only emit the plugins notice after install if none are present

dfreeman:
Hi! I'm in the process of updating our internal app blueprint at work to include some deploy-related stuff, and it now installs ember-cli-deploy and a plugin pack in one go. Given that, I thought it might be nice to avoid emitting the notice about needing plugins if any were already present.

It doesn't look like the blueprint has any test coverage today, so I didn't add anything to capture this change. If you feel it's worth testing, I'm happy to go back and set something up.

### DIFF
--- a/blueprints/ember-cli-deploy/index.js
+++ b/blueprints/ember-cli-deploy/index.js
@@ -9,6 +9,14 @@ module.exports = {
     // to us
   },
   afterInstall: function() {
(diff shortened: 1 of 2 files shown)


duizendnegen:
As some logic is added here now, would be great if you could create some simple tests :)
Thanks for the contribution!

ghedamat:
@dfreeman thanks!

ghedamat:
@dfreeman published!

dfreeman:
@ghedamat awesome, thanks!

PR #491, opened 2017-09-13
[DOC] Updates the signature for configuration properties as functions

xcambar:
## What Changed & Why
The doc at http://ember-cli-deploy.com/docs/v1.0.x/configuration/#usage states that:

> Individual config properties can be pure values or functions that receive the context object and return a value.
>```
>// config/deploy.js
>module.export = function(deployTarget){
>  var ENV = {
>    redis: {
>      revisionKey: function(context) {
>        return context.deployment.tag;
>      }
>    }
>  }
>};
>```

While the code says differently: https://github.com/ember-cli-deploy/ember-cli-deploy-plugin/blob/master/index.js#L68-L70

This PR amends the documentation to reflect the current behaviour.

## PR Checklist
- [ ] Add tests
- [ ] Add documentation
- [x] Prefix documentation-only commits with [DOC]


xcambar:
Oh no... I missed the `.call`.

So this is what coding during a paternity leave looks like 😂

xcambar:
I've amended to only edit the example.
Let's 🤞 that I didn't miss anything this time :)

achambers:
I'm not sure the proposed change is actually accurate. I need to refamiliarise myself with the code and come back to this with more context though.

PR #507, opened 2018-09-17
[DOC] Plugin pack article typo

olleolleolle:
## What Changed & Why

  - The name of a plugin now uses a dash, not an equals sign


lukemelia:
Thanks @olleolleolle. You're on a roll!

PR #517, opened 2020-03-30
add --active-only option to list command

kellyselden:
We need a way to get the active revision, no matter how old it is. I don't think this is possible now, because you would need to know the amount beforehand.


lukemelia:
@kellyselden I understand the need and this looks fine to me. In order to be merged, this PR should include:

- [ ] Documentation update (at least here: https://github.com/ember-cli-deploy/ember-cli-deploy/blob/4866e3aeaaf51c7623acf3c0e1b31310d6d9b4e5/docs/docs/v1.0.x/deploying-your-app.md#listing)

We'll also need to:

- [ ] Resolve unrelated travis build issues

